PR #17, opened 2016-03-09
Fix string conversion of timestamps containing microseconds

patrislav1:
ps_fetch_datetime() does not work for DATETIME(n) fields where n is between 1 and 5.

for example, when DATETIME(3) is used for millisecond resolution, the current version sprintf()s the first 3 digits of the microsecond field, but this yields a wrong output, when there are leading zeroes (see examples below).

Test values, type DATETIME(3):
2016-03-09 07:51:49.000
2016-03-09 07:51:49.001
2016-03-09 07:51:49.010

ps_fetch_datetime() output:
2016-03-09 07:51:49
2016-03-09 07:51:49.1000
2016-03-09 07:51:49.10000


SergMariaDB (changes requested):
Good day @patrislav1,

Please, consider rebasing to the latest HEAD.

Thanks

9EOR9:
Already fixed.

PR #18, opened 2016-04-27
Use PROJECT_SOURCE_DIR instead of CMAKE_SOURCE_DIR

GeorgyKirichenko:
If we use PROJECT_SOURCE_DIR and PROJECT_BINARY_DIR in cmake files then we can use the library as subdirectory in another cmake project.


vaintroub:
Thank you!

PR #19, opened 2016-06-13
Connector c 2.2

counterpoint:
MaxScale uses Connector C to make connections to backend database for monitoring and user authentication (actual proxy services implement the MySQL protocol and do not use Connector C). With the introduction of back end SSL, tests have been done where a call is made to mysql_ssl_set() prior to the call to mysql_real_connect(). Unfortunately, in some cases, the call to mysql_real_connect() then gives an error return, which analyses to "Unknown SSL error". This happens with CentOS 6 and 7 and with Debian 7 and 8. It does not happen with Ubuntu 14.04 or Ubuntu 16.04.


9EOR9:
Closing this issue, since MaxScale doesn't use C/C 2.x anymore.

PR #20, opened 2016-08-11
Prevent crash on uninitialized variable

KhArtNJava:
Hello.
This small PR is to fix crash on referenced, but uninitialized variable. This prevents crash of the code
if (mysql_query(con, "INSERT INTO `test`.`t1` (`Qwe`) VALUES ('Привет');")) {
        finish_with_error(con);
    }
on line 1939 of this file.


9EOR9:
This was fixed in 3.0 rewrite

PR #22, opened 2016-09-03
Add support for TLS SNI during SSL handshake.

FGasper:
This currently is not implemented on any MySQL-compatible servers (?), but there’s no reason not to include it in the client, as it’s basically “free” and, once a server implements SNI, adds a useful feature.


FGasper:
I do have a little Perl test server for this, if that’s desired.

PR #23, opened 2017-01-18
Add io wait callback for blockin io operations.

GeorgyKirichenko:
Add a io wait callback that will be called instead of base io witing
handlers.

We use mariadb-connector-c for our mariadb/mysql connector [https://github.com/tarantool/mysql](url) for tarantool nosql database [https://github.com/tarantool/tarantool](url).

Mariadb connector uses own internal fiber implementation for async operations and need to special wrap for async api function (_start/_cont), but tarantool already has it own fibers.
I added one option for mysql connection that allowed callback from pvio objects for io polling, in this case standart api calls can be used asynchronously without wrapping, because we will need to call some more methods that already exist in future (ex. for replication from mariadb databases).

Little example can be found at our git repo [https://github.com/tarantool/mysql/blob/master/mysql/driver.c](url)

### DIFF
--- a/include/mysql.h
+++ b/include/mysql.h
@@ -208,6 +208,7 @@ extern const char *SQLSTATE_UNKNOWN;
     /* MariaDB specific */
     MYSQL_PROGRESS_CALLBACK=5999,
     MYSQL_OPT_NONBLOCK,
+    MYSQL_OPT_IO_WAIT,
     /* MariaDB Connector/C specific */
     MYSQL_DATABASE_DRIVER=7000,
     MARIADB_OPT_SSL_FP,             /* deprecated, use MARIADB_OPT_TLS_PEER_FP instead */
@@ -316,6 +317,7 @@ struct st_mysql_options {
     int (*local_infile_error)(void *, char *, unsigned int);
     void *local_infile_userdata;
     struct st_mysql_options_extension *extension;
+    int (*io_wait)(my_socket handle, my_bool is_read, int timeout);
 };
 
   typedef struct st_mysql {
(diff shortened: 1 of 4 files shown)


include/mysql.h
```diff
@@ -208,6 +208,7 @@ extern const char *SQLSTATE_UNKNOWN;
     /* MariaDB specific */
     MYSQL_PROGRESS_CALLBACK=5999,
     MYSQL_OPT_NONBLOCK,
+    MYSQL_OPT_IO_WAIT,
```
9EOR9: New options (which are not supported by MySQL server should be prefixed with MariaDB and should be added to the end of enumeration.

include/mysql.h
```diff
@@ -316,6 +317,7 @@ struct st_mysql_options {
     int (*local_infile_error)(void *, char *, unsigned int);
     void *local_infile_userdata;
     struct st_mysql_options_extension *extension;
+    int (*io_wait)(my_socket handle, my_bool is_read, int timeout);
```
9EOR9: This should be added to mysql->options.extension to avoid ABI breakage.

libmariadb/mariadb_lib.c
```diff
@@ -2716,6 +2716,9 @@ mysql_optionsv(MYSQL *mysql,enum mysql_option option, ...)
       }
     mysql->options.extension->async_context= ctxt;
     break;
+  case MYSQL_OPT_IO_WAIT:
+    mysql->options.io_wait = (int(*)(my_socket, bool, int))arg1;
```
9EOR9: callback should be registered in mysql->options.extension->io_wait

9EOR9:
Replaced by PR #71

PR #25, opened 2017-04-26
Update ma_schannel.c to fix compile error

RavenX8:
Fixed compile error due to an incorrect variable name was used.


RavenX8:
Fixed by https://github.com/MariaDB/mariadb-connector-c/commit/b4681a225801381b52d8f831e3914c9bc6a835be

PR #29, opened 2017-07-17
mysql_ssl_set should return 1 if not use tls

KevinIsSun:
If I don't use openssl to encrypt, this method will not be expected return ok.


9EOR9:
Hi,

why do you want to change this? It would break compatibility. Please note that also libmysql returns 0 if not configured with TLS: https://github.com/MariaDB/server/blob/10.1/sql-common/client.c#L1667

9EOR9:
No feedback ->closing request

PR #32, opened 2017-11-12
fix build when libiconv* symbols are exposed.

devnexen:


9EOR9:
Usage of libiconv is disabled by default and will be removed in future releases, so closing this issue.

PR #34, opened 2017-11-12
Change win32 header references to lower-case as capitalization breaks…

omega-software:
… case-sensitive platforms (found on MinGW-w64)

### DIFF
--- a/include/ma_global.h
+++ b/include/ma_global.h
@@ -23,7 +23,7 @@
 
 #ifdef _WIN32
 #include <winsock2.h>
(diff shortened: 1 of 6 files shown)


xx-shitai-xx:
This one would be nice, I have issues with this as well

omega-software:
Hi
This was apparently fixed in commit 821185c20cc845d70e9f5760243148d6229c4ebc

9EOR9:
Fixed (CONC-387)

PR #35, opened 2017-11-12
_GNU_SOURCE isn't available on all GCC platforms (like MinGW-w64)

omega-software:
MinGW-w64 uses GCC but _GNU_SOURCE can't be enabled there as some code uses this macro to enable GNU/Linux facilities like program_invocation_short_name (see libmariadb/mariadb_lib.c).


9EOR9:
_GNU_SOURCE definition was removed.

PR #36, opened 2017-11-12
Fix plugin configuration as either dynamic or built-in

omega-software:
An interface to	configure plugins as either dynamic (DYNAMIC) or built-in (STATIC) is defined in https://mariadb.com/kb/en/library/configuration-settings-for-building-connectorc/.

However this interface isn't actually working because calls to REGISTER_PLUGIN macro in cmake/plugins.cmake override whatever setting is defined by the user with their own.

With this change pre-existing ${name}_PLUGIN_TYPE settings take precedence.


9EOR9:
Closing this issue, since plugin configuration changed in 3.1

PR #37, opened 2017-11-14
Winxp and ws2003

omega-software:
This allows MariaDB Connector/C to be executable again on Windows XP and Windows Server 2003.

Incidentally, it is also needed to build on MinGW-w64 (along with pull requests 34 and 35), until their API is updated.

### DIFF
--- a/libmariadb/ma_context.c
+++ b/libmariadb/ma_context.c
@@ -683,7 +683,15 @@ my_context_spawn(struct my_context *c, void (*f)(void *), void *d)
 int
 my_context_continue(struct my_context *c)
 {
-  void *current_fiber=  IsThreadAFiber() ? GetCurrentFiber() : ConvertThreadToFiber(c);
+  /*
+    This seems to be a common trick to run ConvertThreadToFiber() only on the
+    first occurence in a thread, in a way that works on multiple Windows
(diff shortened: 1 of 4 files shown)


9EOR9:
Why do you want to support eoled platforms like WindowsXP? Last time when I played with MinGW and C/C 3.0 there were lot of other things which didn't work (e.g. crypto and schannel stuff)

omega-software:
Hi Georg

We still have several clients using our software on WinXP. Unfortunately we can't drop support for the time being.

We fixed some other MinGW issues you mention (see https://github.com/MariaDB/mariadb-connector-c/pull/35 and https://github.com/MariaDB/mariadb-connector-c/pull/34).

As for crypto we didn't find any problems (but we use openssl instead of schannel).

omega-software:
Hi, I just wanted to note that our next point release includes MinGW-based MariaDB builds. The MySQL functionality passed QA and will be released in the next few days. It makes a reasonably complete use of crypto features (client verification of server CA, and REQUIRE_X509 directive forcing cert-based auth in the client)

9EOR9:
Closing this issue, since we don't support eoled windows operating systems (and don't have the infrastructure to test on these platforms)

PR #38, opened 2017-12-28
Add support for SESSION_TRACK_GTIDS , CONC-302

renecannao:
Add support for SESSION_TRACK_GTIDS .
Tested with MySQL 5.7 , as MariaDB currently doesn't support it.

https://jira.mariadb.org/browse/CONC-302


9EOR9:
Thanks for your contribution!

Unfortunately we can't accept this pull request without a test. You will find example of session_track tests in unittests/libmariadb/connection.c e.g int test_mdev14647.

9EOR9:
Added in 3.1.10

PR #40, opened 2018-02-17
10.2 server

9EOR9:
Travis and Appveyor test fixes


libmariadb/mariadb_lib.c
```diff
@@ -2653,26 +2657,7 @@ mysql_optionsv(MYSQL *mysql,enum mysql_option option, ...)
...
-        _splitpath(module_filename,NULL, NULL, appname, NULL);
-      }
-#else
-      const char * appname = "";
-#endif
-      OPT_SET_VALUE_STR(&mysql->options, my_cnf_group, appname);
-      break;
-    }
-    OPT_SET_VALUE_STR(&mysql->options, my_cnf_group, (char *)arg1);
+    OPT_SET_VALUE_STR(&mysql->options, my_cnf_group, arg1 ? (char *)arg1 : "");
```
vuvova: Why? I think it was an important feature, contributing to the standard expected behavior of all MariaDB clients

libmariadb/mariadb_lib.c
```diff
@@ -2537,7 +2537,11 @@ mysql_stat(MYSQL *mysql)
 int STDCALL
 mysql_ping(MYSQL *mysql)
 {
-  return ma_simple_command(mysql, COM_PING,0,0,0,0);
+  int rc;
+  rc= ma_simple_command(mysql, COM_PING, 0, 0, 0, 0);
+  if (rc && mysql->options.reconnect)
+    rc= ma_simple_command(mysql, COM_PING, 0, 0, 0, 0);
+  return rc;
 }
```
vuvova: Hmm, perhaps you can reconnect only on certain errors? Like, no need to reconnect if the first ping got "access denied"

libmariadb/mariadb_lib.c
9EOR9: - program name detection failed on several platforms
- In libmysql the default configuration groups will be read if you specify NULL or ""(empty string), but not the programname group.

libmariadb/mariadb_lib.c
9EOR9: Ok, maybe only test if error code was CR_SERVER_LOST ?! (like it was before)

vuvova: @9E0R9 yes. Why did you remove that (your commit had no explanation for removal)?

@vaintroub yes, when using one-time passwords (google authenticator, rsa securid, etc) one cannot reconnect using old credentials.

9EOR9: I think we should check these error codes:

CR_SERVER_GONE_ERROR (2006)
CR_SERVER_LOST (2013)

libmariadb/mariadb_lib.c
vuvova: Well, this is exactly the point. the "standard expected behavior" is to read the default groups and the programname group. Now there's no way to do it. If a client specifies "" or NULL, C/C won't read the programname group. If the client specifies "programname" explicitly, C/C won't read default groups. So no way to get the standard behavior.

On what platforms did the program name detection fail?

9EOR9: The current behavior of Connector/C and libmysql is that configuration files will be read if either config_file or config_group was specified:
```
/* use default options */
  if (mysql->options.my_cnf_file || mysql->options.my_cnf_group)
```
So the initial behavior for mysql_options(mysql, MYSQL_OPT_READ_DEFAULT_GROUP, "") was reading default configuration files with default groups, but not the program name.

Please note that also the my.cnf parser is very basic and will fail, e.g. if your application name has special chars, e.g. "]client"

9EOR9: If I remember correctly, program name wasn't detected on Solaris (this was once fixed in master branch), but also when crosscompiling for android and arm.

PR #42, opened 2018-03-09
Fix the deactivation of gssapi with -DAUTH_GSSAPI_PLUGIN_TYPE=OFF

Thomas-Barbier-1A:


grooverdan:
It looks like this entire `IF` block should be removed as there is a include `cmake/FindGSSAPI.cmake` furhter down in this file that has all the necessary protections against inclusion in undesired/unsupported cases.

PR #43, opened 2018-03-25
MDEV-15655: Add Linux abstract socket support

grooverdan:
Consistent with the server implementation extend the
client connection via unix socket to use an abstract
socket where the start of the path is '@'.

When this occurs the remainder of the name is the
abstract socket socket name. This is consistent with the
approached used by systemd in socket activation.


grooverdan:
Thanks @9EOR9

PR #45, opened 2018-04-14
fixed include

alexroot-:
With this fix one can add mariadb-client as cmake subdirectory in another project.


9EOR9:
For using Connector/C as a subproject please check code from server: https://github.com/MariaDB/server/blob/10.3/cmake/mariadb_connector_c.cmake

PR #49, opened 2018-05-07
CONC-329: my_bool: change to signed char

grooverdan:
Prevents legitimate compile warning on Power and other
architectures that define char to be unsigned (which is allowed by C
standard). In this case the SOCKET_ERROR (-1) is never true.

plugins/pvio/pvio_socket.c:763:42: warning: comparison of constant -1 with expression of type 'my_bool' (aka 'char') is always
      false [-Wtautological-constant-out-of-range-compare]
    if (pvio_socket_blocking(pvio, 1, 0) == SOCKET_ERROR)
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ^  ~~~~~~~~~~~~
plugins/pvio/pvio_socket.c:875:46: warning: comparison of constant -1 with expression of type 'my_bool' (aka 'char') is always
      false [-Wtautological-constant-out-of-range-compare]
        if (pvio_socket_blocking(pvio, 0, 0) == SOCKET_ERROR)
            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ^  ~~~~~~~~~~~~
plugins/pvio/pvio_socket.c:907:42: warning: comparison of constant -1 with expression of type 'my_bool' (aka 'char') is always
      false [-Wtautological-constant-out-of-range-compare]
    if (pvio_socket_blocking(pvio, 1, 0) == SOCKET_ERROR)

I submit this under the MCA.

### DIFF
--- a/include/ma_global.h
+++ b/include/ma_global.h
@@ -714,7 +714,7 @@ typedef short		int15;	/* Most effective integer 0 <= x <= 32767 */
 typedef char		*my_string; /* String of characters */
 typedef unsigned long	size_s; /* Size of strings (In string-funcs) */
 typedef int		myf;	/* Type of MyFlags in my_funcs */
-typedef char		my_bool; /* Small bool */
+typedef signed char	my_bool; /* Small bool */
 typedef unsigned long long my_ulonglong;
 #if !defined(bool) && !defined(bool_defined) && (!defined(HAVE_BOOL) || !defined(__cplusplus))
 typedef char		bool;	/* Ordinary boolean values 0 1 */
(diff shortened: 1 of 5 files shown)


grooverdan:
consistent warnings appear in server build http://buildbot.askmonty.org/buildbot/builders/p8-suse12-rpm/builds/4783/steps/compile/logs/warnings%20%286%29

9EOR9:
Hi,

can you please rebase your commits to a single commit?

grooverdan:
done.

I'm still struggling to work out the mappings between connector-c branches and which branch I should be targeting.

9EOR9:
Hi,

wouldn't it be better to change only the internal blocking function from my_bool to signed char?
my_bool is part of the api (it's also defined as char in 10.1 and MySQL) so I'm not very happy to change it to signed char.

grooverdan:
Alternate implementation in #66 used instead.

PR #51, opened 2018-05-10
MDEV-15655: abstract socket support - fix length(2)

grooverdan:
Thanks to Sergey Vojtovich:

On OS X sockaddr_un is defined as:

struct sockaddr_un
{
  u_char sun_len;
  u_char sun_family;
  char  sun_path[104];
};

There is a comment in man 7 unix (on linux):

On Linux, the above offsetof() expression equates to the same value as sizeof(sa_family_t), but some other implementations include other fields before sun_path, so the offsetof() expression more portably describes the size of the address structure.

As such use the offsetof for Linux and use the previous sizeof(UNIXaddr)
for non-unix platforms as that's what worked before and they don't
support abstract sockets so there's no compatibility problem..


grooverdan:
thanks @vaintroub, #49 is a bit more important. Hope I've targeted the right branch there.

PR #54, opened 2018-05-24
Fix: double free()

FaramosCZ:
The same free() will be called on line 3609 anyway, once send by goto from line 3621 to 3608

### DIFF
--- a/libmariadb/mariadb_dyncol.c
+++ b/libmariadb/mariadb_dyncol.c
@@ -3612,7 +3612,6 @@ end:
 create_new_string:
   /* There is no columns from before, so let's just add the new ones */
   rc= ER_DYNCOL_OK;
-  free(alloc_plan);
   if (not_null != 0)
     rc= dynamic_column_create_many_internal_fmt(str, add_column_count,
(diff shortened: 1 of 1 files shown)


FaramosCZ:
Coverity scan says:

```
Error: USE_AFTER_FREE (CWE-825):
mariadb-connector-c-3.0.4-src/libmariadb/mariadb_dyncol.c:3351: alias: Assigning: "alloc_plan" = "plan". Now both point to the same storage.
mariadb-connector-c-3.0.4-src/libmariadb/mariadb_dyncol.c:3615: freed_arg: "free" frees "alloc_plan".
mariadb-connector-c-3.0.4-src/libmariadb/mariadb_dyncol.c:3609: double_free: Calling "free" frees pointer "alloc_plan" which has already been freed.
# 3607|                                        convert);
# 3608|   end:
# 3609|->   free(alloc_plan);
# 3610|     return rc;
# 3611|   

```

9EOR9:
Thanks for your contribution!

Sanja, can you please check if the dynamic column code in server is also affected?!

PR #66, opened 2018-07-17
CONC-329: change pvio_*_blocking to return int to accomidate SOCKET_ERROR(-1)

grooverdan:
POWER and other architectures that define char(as my_bool) to be unsigned (as the C
standard leaves this undefined). This resulted in error branches being
unreachabe as indicated by the below compile warnings.

plugins/pvio/pvio_socket.c:763:42: warning: comparison of constant -1 with expression of type 'my_bool' (aka 'char') is always
      false [-Wtautological-constant-out-of-range-compare]
    if (pvio_socket_blocking(pvio, 1, 0) == SOCKET_ERROR)
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ^  ~~~~~~~~~~~~
plugins/pvio/pvio_socket.c:875:46: warning: comparison of constant -1 with expression of type 'my_bool' (aka 'char') is always
      false [-Wtautological-constant-out-of-range-compare]
        if (pvio_socket_blocking(pvio, 0, 0) == SOCKET_ERROR)
            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ^  ~~~~~~~~~~~~
plugins/pvio/pvio_socket.c:907:42: warning: comparison of constant -1 with expression of type 'my_bool' (aka 'char') is always
      false [-Wtautological-constant-out-of-range-compare]
    if (pvio_socket_blocking(pvio, 1, 0) == SOCKET_ERROR)

ma_hext2int: signed char - prevent compiler errors when char is unsigned.

libmariadb/ma_tls.c:169:31: warning: comparison of constant -1 with expression of type 'char' is always false
      [-Wtautological-constant-out-of-range-compare]
    if ((d1 = ma_hex2int(*p)) == - 1 ||
        ~~~~~~~~~~~~~~~~~~~~~ ^  ~~~
libmariadb/ma_tls.c:170:35: warning: comparison of constant -1 with expression of type 'char' is always false
      [-Wtautological-constant-out-of-range-compare]
        (d2 = ma_hex2int(*(p+1))) == -1 ||
        ~~~~~~~~~~~~~~~~~~~~~~~~~ ^  ~~

To fix this all the pvio_*_blocking functions have been changed to use
int as a return value.

Other my_bool/char differences fixed:

mariadb_dyncol_val_str: fix prototype to use char - like implemented function.

unittest: bind.is_null is my_bool* so we use a my_bool.

I submit this under the MCA.

### DIFF
--- a/include/ma_pvio.h
+++ b/include/ma_pvio.h
@@ -98,7 +98,7 @@ struct st_ma_pvio_methods
   ssize_t (*write)(MARIADB_PVIO *pvio, const uchar *buffer, size_t length);
   ssize_t (*async_write)(MARIADB_PVIO *pvio, const uchar *buffer, size_t length);
   int (*wait_io_or_timeout)(MARIADB_PVIO *pvio, my_bool is_read, int timeout);
-  my_bool (*blocking)(MARIADB_PVIO *pvio, my_bool value, my_bool *old_value);
+  int (*blocking)(MARIADB_PVIO *pvio, my_bool value, my_bool *old_value);
   my_bool (*connect)(MARIADB_PVIO *pvio, MA_PVIO_CINFO *cinfo);
   my_bool (*close)(MARIADB_PVIO *pvio);
   int (*fast_send)(MARIADB_PVIO *pvio);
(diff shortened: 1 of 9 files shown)


grooverdan:
thanks @9EOR9

PR #69, opened 2018-09-04
CMake: set policy CMP0074 to NEW to fix custom package paths for CMake 3.12 and above

davidebeatrici:
From https://cmake.org/cmake/help/git-stage/policy/CMP0074.html:
>[find_package()](https://cmake.org/cmake/help/git-stage/command/find_package.html#command:find_package) uses `<PackageName>_ROOT` variables.
>
>In CMake 3.12 and above the [`find_package(<PackageName>)`](https://cmake.org/cmake/help/git-stage/command/find_package.html#command:find_package) command now searches prefixes specified by the [`<PackageName>_ROOT`](https://cmake.org/cmake/help/git-stage/variable/PackageName_ROOT.html#variable:%3CPackageName%3E_ROOT) CMake variable and the [`<PackageName>_ROOT`](https://cmake.org/cmake/help/git-stage/variable/PackageName_ROOT.html#variable:%3CPackageName%3E_ROOT) environment variable. Package roots are maintained as a stack so nested calls to all `find_*` commands inside find modules also search the roots as prefixes. This policy provides compatibility with projects that have not been updated to avoid using `<PackageName>_ROOT` variables for other purposes.
>
>The `OLD` behavior for this policy is to ignore `<PackageName>_ROOT` variables. The `NEW` behavior for this policy is to use `<PackageName>_ROOT` variables.
>
>This policy was introduced in CMake version 3.12. CMake version 3.12.20180904-gb694d warns when the policy is not set and uses `OLD` behavior. Use the [cmake_policy()](https://cmake.org/cmake/help/git-stage/command/cmake_policy.html#command:cmake_policy) command to set it to `OLD` or `NEW` explicitly.


davidebeatrici:
@9EOR9 As far as I can see, this was never addressed:

https://github.com/mariadb-corporation/mariadb-connector-c/blob/52d0a38ed15f62906206d77b675079fc159cec7e/CMakeLists.txt#L5-L16

PR #71, opened 2018-10-13
Add IOWait callback for blocking IO operations

bigbes:
Add a io wait callback that will be called instead of base io writing handlers.

We use mariadb-connector-c for our mariadb/mysql connector https://github.com/tarantool/mysql for tarantool nosql database https://github.com/tarantool/tarantool.

Mariadb connector uses own internal fiber implementation for async operations and need to special wrap for async api function (_start/_cont), but tarantool already has it own fibers.
I added one option for mysql connection that allowed callback from pvio objects for io polling, in this case standart api calls can be used asynchronously without wrapping, because we will need to call some more methods that already exist in future (ex. for replication from mariadb databases).

Little example can be found at our git repo https://github.com/tarantool/mysql/blob/master/mysql/driver.c

This PR fixes reviews in https://github.com/MariaDB/mariadb-connector-c/pull/23. Sorry, I was forced to open new PR, since I haven't got access to old repo. I've fixed @9EOR9 remarks from old PR.


bigbes:
Rebased onto latest branch, build fails with unrelated to patch problems.

9EOR9:
- Filed [Jira TASK](https://jira.mariadb.org/browse/CONC-377) 
- Applied part of the patch to 3.1 branch

The patch for named pipe doesn't work, since wait_io was reworked recently by vvaintroub.

PR #73, opened 2018-10-30
CONC-369 libmariadb doesn't compile with WITH_TSAN cmake option

kevgs:
Do not "-Wl,--no-undefined" if WITH_TSAN enabled.

I'm contributing this new code of the whole pull request, including one or several files that are either new files or modified ones, under the BSD-new license.


9EOR9:
This was already fixed in 3.0.7

kevgs:
Oh. I really can see it now in `10.3` and `10.4` branches. Sorry for bothering you: I messed up connector branches.

PR #74, opened 2018-11-21
Use lower case in included files

mlocati:
I'm building the Windows DLLs `auth_gssapi_client.dll` and `caching_sha2_password.dll` under Ubuntu with MXE.

The compilation fails with errors like these:
```
include/ma_global.h:26:21: fatal error: Windows.h: No such file or directory
 #include <Windows.h>
                     ^
compilation terminated.
```

The problem is that mingw has `windows.h` and not `Windows.h` (Linux file system is case-sensitive).
In order to fix it, we need lower-case includes.


9EOR9:
Fixed in 3.0 branch (3.0.8)

PR #75, opened 2018-11-21
Define missing potentially defines in sspi_errmsg.c

mlocati:
I'm building the Windows DLLs auth_gssapi_client.dll and caching_sha2_password.dll under Ubuntu with MXE.

The compilation fails with errors like these:

```
plugins/auth/sspi_errmsg.c:127:10: error: 'SEC_E_INVALID_PARAMETER' undeclared here (not in a function)
   ERRSYM(SEC_E_INVALID_PARAMETER),
          ^
```

The problem is that mingw has a `winerror.h` file that's missing some definitions.


9EOR9:
Fixed in 3.0 branch (3.0.8)

mlocati:
@9EOR9 any reason why I don't see my commits in https://github.com/MariaDB/mariadb-connector-c/commits/3.0 ?

9EOR9:
I combined them instead of commit several change sets which affect the same issue.

mlocati:
I don't care that much, but it doesn't seems to me it's correct to loose the commit authors...

PR #76, opened 2018-11-21
Cast constant string as char *

mlocati:
This fixes the following compilation warning:

```
plugins/auth/sspi_client.c:82:10: error: assignment discards 'const' qualifier from pointer target type [-Werror=discarded-qualifiers]
     mech = "Kerberos";
          ^
```


9EOR9:
Applied to 3.0 branch (3.0.8)

PR #78, opened 2018-11-21
Add -Wno-unknown-pragmas to GCC flags

mlocati:
gcc doesn't recognize `#pragma comment`


9EOR9:
Fixed in 3.0 branch (3.0.8)

PR #79, opened 2018-11-21
Add bcrypt to the libraries for caching_sha2_password

mlocati:


9EOR9:
Bcrypt should be added to the caching libs only if SSL option was set to schannel. You can also build Conector/C with OpenSSL or GnuTLS support on Windows.
Corrected/fixed patch applied to 3.0 branch

PR #80, opened 2018-11-21
Fix type of pcbStructInfo param of CryptDecodeObjectEx

mlocati:
The `pcbStructInfo` parameter of `CryptDecodeObjectEx` is `DWORD*`, not `int*`


9EOR9:
Patch applied to 3.0 branch

PR #81, opened 2018-11-21
Fix type of pcbResult param of BCryptEncrypt

mlocati:
The `pcbResult` parameter of `BCryptEncrypt` is `ULONG*`, not `unsigned int*`


9EOR9:
Patch applied to 3.0 branch

PR #82, opened 2018-11-21
Fix type of pbInput param of BCryptEncrypt

mlocati:
The `pbInput` parameter of `BCryptEncrypt` is `PUCHAR` (aka `unsigned char *`), not `char *`


9EOR9:
Patch applied to 3.0 branch

PR #83, opened 2018-11-21
Consistent type result of ternary operator in auth_caching_sha2_client

mlocati:
We have that `filebuffer` is a `char*` and `packet` is an `unsigned char *`.
Let's make the ternary operator always return the same type (`char*`, the one expected by `ma_load_pem`).


9EOR9:
Patch applied to 3.0 branch

PR #88, opened 2018-12-11
Compile for Windows Vista and later

mlocati:
Fix this linking error:
```
undefined reference to `CryptImportPublicKeyInfoEx2'
```


9EOR9:
Windos build works ok for all supported Windows platforms. There is no need to support eoled version of windows, we also don't have testing machines for eoled windows versions, same for Appveyor and Travis.

PR #90, opened 2018-12-11
Skip PDB handling if compiler is gcc

mlocati:
gcc does not generate PDB files


gleichdick:
Ran into the same Issue and was about to make a new PR. Happy to see the fix upstream :+1:

cmake/install_plugins.cmake
```diff
@@ -11,6 +11,8 @@ MACRO(INSTALL_PLUGIN name binary_dir)
   INSTALL(TARGETS ${name} COMPONENT ClientPlugins DESTINATION ${INSTALL_PLUGINDIR})
   IF(WIN32)
     FILE(APPEND ${CC_BINARY_DIR}/win/packaging/plugin.conf "<File Id=\"${name}.dll\" Name=\"${name}.dll\" DiskId=\"1\" Source=\"${binary_dir}/${CMAKE_BUILD_TYPE}/${name}.dll\"/>\n")
```
vuvova, line 13: do you need plugin.conf file at all? it seems to be some MSI installer stuff, can you even build an MSI installer with mingw?

PR #93, opened 2019-01-02
Fix table map event processing

markus456:
The table name is also null-terminated so the length plus one byte needs
to be skipped.

If the table map event is not used immediately after the event is read,
the column types would point to possibly freed memory. To avoid this,
memory should be allocated for it.


markus456:
Opened another pull request for the 3.1 branch.

PR #100, opened 2019-01-18
Add binlog checksum support

markus456:
The source code was built into the library but the header was not
installed.


markus456:
Added support for binlog checksums which fixes the binlog names in ROTATE_EVENTs as well as the data in ROWS events.

markus456:
@9E0R9 could you review this?

PR #101, opened 2019-01-20
Fix my.cnf "host" option

bugfood:
It has been possible in the past to have a my.cnf file containing:

```
[example]
host = db.example.com
user = example_user
password = example_password
```
Clients using the mysql or mariadb library (eg perl's DBD::mysql in my case) could be directed to use the `example` group in order to fetch all the necessary settings.

Recently, this feature stopped working--the client would attempt to connect to the local socket rather than the specified host.

This PR fixes that and add some well-intended and hopefully-correct error checking. If the error checking is not desired, than I can submit a simple version without it.

Thanks,
Corey

### DIFF
--- a/libmariadb/mariadb_lib.c
+++ b/libmariadb/mariadb_lib.c
@@ -1206,7 +1206,9 @@ MYSQL *mthd_my_real_connect(MYSQL *mysql, const char *host, const char *user,
   if (!host || !host[0])
     host = mysql->options.host;
 
-  ma_set_connect_attrs(mysql, host);
+  if (ma_set_connect_attrs(mysql, host))
+    /* client error should be set already */
+    return(NULL);
 
   if (net->pvio)  /* check if we are already connected */
   {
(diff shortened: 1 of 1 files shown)


bugfood:
I accidentally left in a debug printf(); removed now.

bugfood:
As far as I can tell, the test failures are unrelated to my changes.

libmariadb/mariadb_lib.c
```diff
@@ -3019,6 +3019,7 @@ mysql_optionsv(MYSQL *mysql,enum mysql_option option, ...)
     break;
   default:
     va_end(ap);
+    SET_CLIENT_ERROR(mysql, CR_NOT_IMPLEMENTED, SQLSTATE_UNKNOWN, 0);
     return(-1);
```
9EOR9: Better to change the return code from -1 to 1. According to the documentation mysql_options/v will return non zero return code on error, so this will not break anything.

9EOR9:
Last part pushed  into 3.0.10 (mysql_options/v was aready pushed into 3.0.9) - Thanks for your help and contribution.

PR #103, opened 2019-02-22
Fix CONC-390

0x1EE7:
Fix for CONC-301 for some reason seems to be reverted starting from 3.0.4. Could not find much information about why but this PR is to make sure it is not lost.
More info:
https://wiki.strongswan.org/issues/2771
https://jira.mariadb.org/browse/CONC-390


0x1EE7:
@9EOR9 @vuvova Any thoughts on this?

9EOR9:
Fixed in 3.0, this will be merged from 3.0 to 3.1

PR #104, opened 2019-03-04
Fix hardcoded path

FaramosCZ:
On Fedora the client library is named "libmariadb.so", which cofuses the linker which can't find the hardcoded library.


FaramosCZ:
Submitted under MCA

9EOR9:
The examples directory will be removed, it is used by old doc book sources (connector-c-docs github project) and not by Connector/C itself.

9EOR9:
examples directory was removed in 3.1.11

PR #106, opened 2019-03-21
 Reserve the correct amount of space for date/time types 

Zyrin:
This fixes sporadic crashes due to heap corruption


9EOR9:
Connector/C 2.x is not supported anymore.
Also your patch is wrong: The correct lengt is 12 for fractional seconds, while without fractional seconds it must be 8 (not 7). This was fixed Long time ago in C/C 3.0

PR #107, opened 2019-03-21
Reserve the correct amount of space for date/time types

Zyrin:
This fixes a crash, when more then 1024 Bytes are required, but the incorrect amount of memory was reserved, leading to a heap corruption.

### DIFF
--- a/libmariadb/mariadb_stmt.c
+++ b/libmariadb/mariadb_stmt.c
@@ -790,8 +790,16 @@ unsigned char* mysql_stmt_execute_generate_simple_request(MYSQL_STMT *stmt, size
         case MYSQL_TYPE_ENUM:
         case MYSQL_TYPE_BIT:
         case MYSQL_TYPE_SET:
-          size+= 5; /* max 8 bytes for size */
+          size+= 9; /* max 9 bytes for size */
           size+= (size_t)ma_get_length(stmt, i, 0);
+          break;
+        case MYSQL_TYPE_DATE:
+        case MYSQL_TYPE_TIMESTAMP:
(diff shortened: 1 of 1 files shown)


libmariadb/mariadb_stmt.c
```diff
@@ -790,9 +790,17 @@ unsigned char* mysql_stmt_execute_generate_simple_request(MYSQL_STMT *stmt, size
         case MYSQL_TYPE_ENUM:
         case MYSQL_TYPE_BIT:
         case MYSQL_TYPE_SET:
-          size+= 5; /* max 8 bytes for size */
+          size+= 9; /* max 9 bytes for size */
           size+= (size_t)ma_get_length(stmt, i, 0);
           break;
+        case MYSQL_TYPE_DATE:
```
9EOR9: These types represent a binary type (MYSQL_TIME), so why do you set the length to the corresponding string representation?

9EOR9:
Can you please also add a test case (unittest/libmysql/ps_bugs.c)?

libmariadb/mariadb_stmt.c
Zyrin: Maybe the used constants are not correct, but those are the constants store_param uses to allocate memory.
I had to handle the Time cases, because otherwise the default case adds "-1" to the size, whitch can lead to a heap corruption in some cases, when space is reallocated.

Zyrin:
I will write one, but that may take a while.

9EOR9:
Fixed by rev. dce4b8a89d090228f5145a36c66a272e897c82df

PR #116, opened 2019-07-05
Revert "MDEV-16383 Add mariadb_config --libmysqld-libs option"

9EOR9:
Reverts MariaDB/mariadb-connector-c#109
I didn't see it's master (which isn't in use anymore)  - can you please add the changes in 3.0 branch? We will merge it in 3.1 afterwards

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -253,13 +253,6 @@ ELSEIF()
   SET(HAVE_THREADS ${CMAKE_USE_PTHREADS})
 ENDIF()
 
-# only let mariadb-config return the embedded server build flags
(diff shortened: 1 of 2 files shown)


ottok:
I wonder what is the status here? We are still carrying this patch in Debian until upstream merges it or implements the equivalent: https://salsa.debian.org/mariadb-team/mariadb-10.4/-/blob/master/debian/patches/928230-mysql_config---libmysqld-libs.patch

ottok:
@9E0R9 I wonder what is the status here?

9EOR9:
Closed- fixed in 3.1 - 3.3

PR #120, opened 2019-09-09
Compatibility with MySQL server 8.0

yurriy:
Fixed caching_sha2_password plugin behaviour when SSL is enabled.
Currently, when password is non-empty and ssl is enabled, MariaDB connector sends a password as a plaintext before Fast Authentication, and server responds with an [error](https://github.com/mysql/mysql-server/blob/8.0/sql/auth/sha2_password.cc#L1003) as it expects the client to perform Fast Authentication by sending a scramble of size of SHA256.
[Caching SHA2 description](https://dev.mysql.com/doc/dev/mysql-server/latest/page_caching_sha2_authentication_exchanges.html#sect_caching_sha2_definition)


9EOR9:
Thanks for you contribution - merged into 3.1

PR #121, opened 2019-09-26
gssapi: fix include path for Solaris

grooverdan:
tested on:

uname -a
SunOS openindiana 5.11 illumos-b97b1727bc i86pc i386 i86pc

I submit this under the MCA.


grooverdan:
thanks for the quick merge.

PR #122, opened 2019-11-13
Fix run-time issue

rprasanth:
Found this issue while checking in my target, when the library is moved to usr/lib it works

gcc test `mariadb_config --libs --cflags` -o test
./test: error while loading shared libraries: libmariadb.so.3: cannot
open shared object file: No such file or directory

where,
 mariadb_config --libs --cflags
-L/usr/lib/mariadb/ -lmariadb
-I/usr/include/mariadb -I/usr/include/mariadb/mysql


rprasanth:
sorry wrong fix, but the issue exists

PR #124, opened 2020-01-10
[CONC-430] Compilation on AIX

EGuesnet:
This PR improves the behavior of mariadb-conector-c on AIX. Without, it does not compile.

Mariadb-connector-c was tested with mariadb-server (patch not submitted yet); 20 fail on 3640 tests on debug mode, 5 fail on 3339 tests on release mode.

Code compiled with GCC 8.3 (both 32 and 64 bits) on AIX 6.1 and with CMake 3.16.0.

Issue associated: https://jira.mariadb.org/browse/CONC-430.

----
Comment about commit:

Some specific code are no more needed. It breaks the compilation.
Some typedef are not consistant. They are defined in another way on ``/usr/include/sys/inttypes.h``. With "signed", typedef is consistant with system and with comments.
Definition of MSG_DONTWAIT is made on ``libmariadb/mariadb_async.c`` for AIX. It is not enought, I add also it on plugins/pvio/pvio_socket.c file.

Thanks.


EGuesnet:
With this PR, mariadb-connector-c will compile and work on AIX. It is only AIX-specific code, and a correction of compatibility typedef.

Can someone have a look?

grooverdan:
Looks good too me. All either AIX contained or explicitly `signing` char is essential as the C standard doesn't define it ([previous similar fix](https://github.com/mariadb-corporation/mariadb-connector-c/pull/66/files#diff-86111537d7fd2a9803d59a2c0de4e516R164)). `signed short` seems the default and if its already in the header its a good fix to be consistent.

EGuesnet:
Hi,
Anything new here? Does something block this PR?

grooverdan:
@9E0R9 are you ok with this?

mariadb_config/CMakeLists.txt
```diff
@@ -52,6 +52,10 @@ CONFIGURE_FILE(${CMAKE_CURRENT_SOURCE_DIR}/libmariadb.pc.in
 
 ADD_EXECUTABLE(mariadb_config ${CMAKE_CURRENT_BINARY_DIR}/mariadb_config.c)
 
+IF(CMAKE_SYSTEM_NAME MATCHES AIX)
+  TARGET_LINK_LIBRARIES(mariadb_config compat-getopt)
```
grooverdan, line 56: @EGuesnet what provides the compat-getopt library?

libmariadb/CMakeLists.txt
```diff
@@ -309,7 +309,9 @@ ELSE()
   IF(ICONV_INCLUDE_DIR)
      INCLUDE_DIRECTORIES(BEFORE ${ICONV_INCLUDE_DIR}) 
   ENDIF()
-  ADD_DEFINITIONS(-DLIBICONV_PLUG)
+  IF(NOT CMAKE_SYSTEM_NAME MATCHES AIX)
+    ADD_DEFINITIONS(-DLIBICONV_PLUG)
+  ENDIF()
```
grooverdan, line 314: iconv related, but not this change:

I got link errors using the 3.1 branch as updated in libmariadb under mariadb-server
```
Scanning dependencies of target libmariadb
[ 12%] Linking C shared library libmariadb
ld: 0711-317 ERROR: Undefined symbol: .libiconv_open
ld: 0711-317 ERROR: Undefined symbol: .libiconv
ld: 0711-317 ERROR: Undefined symbol: .libiconv_close
ld: 0711-345 Use the -bloadmap or -bnoquiet option to obtain more information.
```

It turns out  `/usr/lib/libiconv.a` doesn't define these symbols,  `/usr/linux/lib/libiconv.a`  does.

I used the ugly hack of
```
--- a/cmake/FindIconv.cmake
+++ b/cmake/FindIconv.cmake
@@ -28,6 +28,9 @@ ELSEIF(APPLE)
                /usr/lib/
                NO_CMAKE_SYSTEM_PATH)
     SET(ICONV_EXTERNAL TRUE)
+ELSEIF(CMAKE_SYSTEM_NAME MATCHES "AIX")
+  SET(ICONV_LIBRARIES -liconv)
+  SET(ICONV_EXTERNAL TRUE)
 ELSE()
   find_library(ICONV_LIBRARIES NAMES iconv libiconv libiconv-2)
   IF(ICONV_LIBRARIES)
```
To work around this. Do you have any ideas on making this more robust?

mariadb_config/CMakeLists.txt
EGuesnet: The getopt_long() function is not available on AIX by default. Opensource teams on AIX have created compatibility libraries to provide some function.

grooverdan: where is it available?

libmariadb/CMakeLists.txt
EGuesnet: We usually use libraries built by opensource teams (in /usr/linux/lib or /opt/freeware/lib), even if they are available on base system (except ssl and crypto). So, I have not faced this issue. Your hack seems to be compatible with both version of libiconv. I do not have idea to make it more robust.

mariadb_config/CMakeLists.txt
EGuesnet: The [lib](http://www.bullfreeware.com/pkg?id=5803), and [devel](http://www.bullfreeware.com/pkg?id=5804) package. They are not available yet on _AIX Toolbox for Linux Applications._

libmariadb/CMakeLists.txt
grooverdan: Thanks. Highlighting it has made a push to remove it as a dependency anyway.

PR #125, opened 2020-01-29
[misc] MariaDB 10.5 testing

rusher:
Adding Appveyor 10.4 and 10.5 testing. 
improving appveyor server DL : testing on standard mirror or archive if not available on mirror. 

Changed travis testing using 10.5 build in place of 10.4


9EOR9:
This is already fixed (together with TLS fixes)

PR #127, opened 2020-02-11
add @CMAKE_SYSROOT@ to include and library paths

ryco-dev:
When cross-compiling MariaDB, system paths are returned by the mysql-config
and mariadb-config tools. Add @CMAKE_SYSROOT@ to the paths to avoid this
issue.

Signed-off-by: Ryan Coe <bluemrp9@gmail.com>


9EOR9:
Issue was merged. Thanks for your contribution!

PR #129, opened 2020-03-04
Fix typo in output string: inital -> initial

ottok:


ottok:
Travis-CI tests seem to fail with an unrelated error
> Error: SSL connection error: self signed certificate

ottok:
Thanks!

PR #131, opened 2020-03-30
Implement CLIENT_DEPRECATE_EOF for mariadb-connector-c

bibstha:
Fix for [CONC-461](https://jira.mariadb.org/browse/CONC-461)

mariadb-connector-c client now sends CLIENT_DEPRECATE_EOF flag to the server.

There are some major changes on how the server responds. Mainly the way packets are sent back from server changes which are documented in https://dev.mysql.com/worklog/task/?id=7766

I will try to add as much comment on the code changes as possible.

Note:

Most of the changes have been studied on the worklog, looking into the packets through wireshark and how mysql client behaves.

### DIFF
--- a/include/ma_common.h
+++ b/include/ma_common.h
@@ -23,6 +23,8 @@
 #include <mysql.h>
 #include <ma_hash.h>
 
+#define MAX_PACKET_LENGTH (256L*256L*256L-1)
+
 enum enum_multi_status {
   COM_MULTI_OFF= 0,
   COM_MULTI_CANCEL,
--- a/libmariadb/mariadb_lib.c
+++ b/libmariadb/mariadb_lib.c
@@ -142,6 +142,7 @@ struct st_mariadb_methods MARIADB_DEFAULT_METHODS;
 #define IS_CONNHDLR_ACTIVE(mysql)\
   ((mysql)->extension && (mysql)->extension->conn_hdlr)
 
+int ma_read_ok_packet(MYSQL *mysql, uchar *pos, ulong length);
 static void end_server(MYSQL *mysql);
 static void mysql_close_memory(MYSQL *mysql);
 void read_user_name(char *name);
@@ -188,10 +189,13 @@ void net_get_error(char *buf, size_t buf_len,
 *****************************************************************************/
 
 ulong
-ma_net_safe_read(MYSQL *mysql)
+ma_net_safe_read(MYSQL *mysql, my_bool *is_data_packet)
 {
   NET *net= &mysql->net;
   ulong len=0;
+
+  if (is_data_packet)
+    *is_data_packet= FALSE;
 
 restart:
   if (net->pvio != 0)
@@ -248,6 +252,38 @@ restart:
     mysql->server_status&= ~SERVER_MORE_RESULTS_EXIST;
 
     return(packet_error);
+  }
+  else
+  {
+    /*
+      Now we have a data packet, unless it is OK packet starting with
+      0xFE - we detect that case below.
+    */
+    if (is_data_packet)
+      *is_data_packet= TRUE;
+    /*
+       For a packet starting with 0xFE detect if it is OK packet or a
+       huge data packet. Note that old servers do not send OK packets
+       starting with 0xFE.
+    */
+    if ((mysql->server_capabilities & CLIENT_DEPRECATE_EOF) &&
+        (net->read_pos[0] == 254))
+    {
+      /* detect huge data packet */
+      if (len > MAX_PACKET_LENGTH)
+        return len;
+      /* otherwise we have OK packet starting with 0xFE */
+      if (is_data_packet)
+        *is_data_packet= FALSE;
+      return len;
+    }
+    /* for old client detect EOF packet */
+    if (!(mysql->server_capabilities & CLIENT_DEPRECATE_EOF) &&
+        (net->read_pos[0] == 254) && (len < 8))
+    {
+      if (is_data_packet)
+        *is_data_packet= FALSE;
+    }
   }
   return len;
 }
@@ -422,7 +458,7 @@ mthd_my_send_cmd(MYSQL *mysql,enum enum_server_command command, const char *arg,
 
   if (!skipp_check)
   {
-    result= ((mysql->packet_length=ma_net_safe_read(mysql)) == packet_error ?
+    result= ((mysql->packet_length=ma_net_safe_read(mysql, NULL)) == packet_error ?
 	     1 : 0);
   }
  end:
@@ -563,12 +599,13 @@ end_server(MYSQL *mysql)
 void mthd_my_skip_result(MYSQL *mysql)
 {
   ulong pkt_len;
+  my_bool is_data_packet;
 
   do {
-    pkt_len= ma_net_safe_read(mysql);
+    pkt_len= ma_net_safe_read(mysql, &is_data_packet);
     if (pkt_len == packet_error)
       break;
-  } while (pkt_len > 8 || mysql->net.read_pos[0] != 254);
+  } while (mysql->net.read_pos[0] == 0 || is_data_packet);
   return;
 }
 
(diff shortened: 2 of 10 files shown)


include/ma_common.h
```diff
@@ -23,6 +23,8 @@
 #include <mysql.h>
 #include <ma_hash.h>
 
+#define MAX_PACKET_LENGTH (256L*256L*256L-1)
```
bibstha, line 26: Moved from ma_net.c from below: https://github.com/mariadb-corporation/mariadb-connector-c/pull/131/files#diff-4ad0e2e8b53cd72422b2c0863655c5e5L42 since it is required in `mariadb_lib.c`

libmariadb/mariadb_lib.c
```diff
@@ -188,11 +189,14 @@ void net_get_error(char *buf, size_t buf_len,
 *****************************************************************************/
 
 ulong
-ma_net_safe_read(MYSQL *mysql)
+ma_net_safe_read(MYSQL *mysql, my_bool *is_data_packet)
```
bibstha, line 192: Similar to https://github.com/mysql/mysql-server/blob/bd9c260cf7026a54d9bf91ce6782cdb4e6a25c71/sql-common/client.c#L1168

libmariadb/mariadb_lib.c
```diff
@@ -849,83 +886,114 @@ static size_t rset_field_offsets[]= {
...
+  @param server_capabilities  protocol capability flags which determine format of
+                              the descriptor
+  @param row            field descriptor
+  @param field          address of MYSQL_FIELD structure to store metadata in.
+
+  @returns 0 on success.
+*/
+
+int
+unpack_field(const MYSQL *mysql, MA_MEM_ROOT *alloc, my_bool default_value,
```
bibstha, line 907: This is extracted as is from `unpack_fields` function below so it can be called directly when reading packet metadata similar to mysql source code [here](https://github.com/mysql/mysql-server/blob/bd9c260cf7026a54d9bf91ce6782cdb4e6a25c71/sql-common/client.c#L2036)

libmariadb/mariadb_lib.c
```diff
@@ -1080,6 +1161,75 @@ int mthd_my_read_one_row(MYSQL *mysql,uint fields,MYSQL_ROW row, ulong *lengths)
...
+ *
+ * With CLIENT_DEPRECATE_EOF now being sent, the response looks like
+ * <FieldCount><Metadata><ResultSet * n><EOF>. The number of metadata to look for
+ * is already available in FieldCount therefore first EOF is not necessary.
+ *
+ * Therefore this method is introduced to handle the new format
+ * and is called by both
+ * mthd_my_read_query_result and mthd_stmt_read_prepare_response
+ */
+MYSQL_FIELD *mthd_my_read_metadata_ex(MYSQL *mysql, MA_MEM_ROOT *alloc,
```
bibstha, line 1175: Similar to [mysql client](https://github.com/mysql/mysql-server/blob/bd9c260cf7026a54d9bf91ce6782cdb4e6a25c71/sql-common/client.c#L2179)

libmariadb/mariadb_lib.c
```diff
@@ -1080,6 +1161,75 @@ int mthd_my_read_one_row(MYSQL *mysql,uint fields,MYSQL_ROW row, ulong *lengths)
...
+    {
+      mysql->warning_count= uint2korr(pos + 1);
+      mysql->server_status= uint2korr(pos + 3);
+    }
+  }
+
+  return result;
+}
+
+MYSQL_FIELD *mthd_my_read_metadata(MYSQL *mysql, ulong field_count, uint field)
```
bibstha, line 1225: The reason mthd_my_read_metadata_ex and mthd_my_read_metadata are two methods: similar to how it is implemented in mysql client. `mthd_my_read_metadata_ex` takes `MA_MEM_ROOT` so it can be called from `mthd_stmt_read_prepare_response`

libmariadb/mariadb_lib.c
```diff
@@ -2298,20 +2447,21 @@ int mthd_my_read_query_result(MYSQL *mysql)
...
   mysql->extra_info= net_field_length_ll(&pos); /* Maybe number of rec */
-  if (!(fields=mysql->methods->db_read_rows(mysql,(MYSQL_FIELD*) 0,
-                                            ma_result_set_rows(mysql))))
-    return(-1);
-  if (!(mysql->fields=unpack_fields(mysql, fields, &mysql->field_alloc,
-				    (uint) field_count, 1)))
-    return(-1);
+
+  if (!(mysql->fields=mthd_my_read_metadata(mysql, field_count,
+          7 + ma_extended_type_info_rows(mysql))))
```
bibstha, line 2458: `ma_result_set_rows` returns 8 or 9, which worked if calling `db_read_rows`. Now that we're reading meta directly from `mthd_my_read_metadata`, by default field = 7. [ref](https://github.com/mysql/mysql-server/blob/bd9c260cf7026a54d9bf91ce6782cdb4e6a25c71/sql-common/client.c#L4990-L4991)

libmariadb/mariadb_lib.c
```diff
@@ -4341,41 +4488,41 @@ struct st_mariadb_api MARIADB_API=
  */
 struct st_mariadb_methods MARIADB_DEFAULT_METHODS = {
   /* open a connection */
-  mthd_my_real_connect,
+  mthd_my_real_connect, // db_connect
```
bibstha, line 4489: Adding the actual names so it's easier to grep.

libmariadb/mariadb_stmt.c
```diff
@@ -274,11 +286,34 @@ int mthd_stmt_read_all_rows(MYSQL_STMT *stmt)
     } else  /* end of stream */
     {
       *pprevious= 0;
-      /* sace status info */
-      p++;
-      stmt->upsert_status.warning_count= stmt->mysql->warning_count= uint2korr(p);
-      p+=2;
-      stmt->upsert_status.server_status= stmt->mysql->server_status= uint2korr(p);
+      /* save status info */
+      if (stmt->mysql->server_capabilities & CLIENT_DEPRECATE_EOF && !is_data_packet)
```
bibstha, line 290: [Ref mysql source](https://github.com/mysql/mysql-server/blob/bd9c260cf7026a54d9bf91ce6782cdb4e6a25c71/libmysql/libmysql.c#L4497)

libmariadb/mariadb_stmt.c
```diff
@@ -334,31 +369,56 @@ static int stmt_cursor_fetch(MYSQL_STMT *stmt, uchar **row)
...
+      in_resultset = 1; // so we do not check for OK pkt with 0 next time in the loop
+
+      if (*pos != 0 && !is_data_packet)
+      {
+        ma_read_ok_packet(mysql, pos + 1, packet_len);
+        goto end;
+      }
     }
-    if (packet_len < 8 && *pos == 254) /* EOF */
+    else
```
bibstha, line 398: The else part is as is as before. Only the upper if conditional is different.

libmariadb/mariadb_stmt.c
```diff
@@ -1869,6 +1925,48 @@ int stmt_read_execute_response(MYSQL_STMT *stmt)
...
+        it responds with binary row we add it to the binary rows result set
+        (the reset of the result set will be read in prepare_to_fetch_result).
+      */
+
+      if ((pkt_len= ma_net_safe_read(mysql, &is_data_packet)) == packet_error)
+        return(1);
+
+      if (is_data_packet)
+      {
+        // TODO: <metadata><row(s)><OK> not supported right now
```
bibstha, line 1961: Note that I'm still unsure how to reproduce it here. Non of the existing tests capture it yet.

libmariadb/mariadb_stmt.c
```diff
@@ -1869,6 +1925,48 @@ int stmt_read_execute_response(MYSQL_STMT *stmt)
   if (!stmt->mysql)
     return(1);
 
+  if ((mysql->server_capabilities & CLIENT_DEPRECATE_EOF))
```
bibstha, line 1928: [Ref](https://github.com/mysql/mysql-server/blob/bd9c260cf7026a54d9bf91ce6782cdb4e6a25c71/libmysql/libmysql.c#L2166)

bibstha:
@9E0R9 do you have some time to look into this?

9EOR9:
Thanks for your pull request - I will review it next week.

9EOR9:
Can you please stash multiple commits into one commit? That makes reviewing much easier. Thans

bibstha:
Done, I squashed them all into a single one.

9EOR9:
From a technical standpoint the code looks more or less ok, but from legal it doesn't:

There is a lot of code (and even comments) just copied from Oracle/MySQL libmysql which is under GPL license and therefore not compatible with LGPL license of MariaDB Connector/C.

bibstha:
@9E0R9 I see, I'll approach the code in a different way than mysql.

ioquatix:
I'm playing around with this feature, what are the benefits of CLIENT_DEPRECATE_EOF?

bibstha:
Hi @ioquatix,
The main benefit for CLIENT_DEPRECATE_EOF is that each mysql response now supports additional information in the form of OK_packets, ie: session tracking information. [More info on session tracking info](https://dev.mysql.com/doc/internals/en/packet-OK_Packet.html)

SessionTracking is not just for pre-defined data, but you can write plugins to send back custom data too. At work, we're using this to send back performance metrics from inside the database.

PS: We also added support for session-tracking to mysql2 gem in https://github.com/brianmario/mysql2/pull/1092 (merged into master, not yet released.)

vaintroub:
The MySQL implementation was done by someone without much experience in MySQL protocol, and is even conceptually broken w.r.t old features. For example, the "stored procedure output parameters" from 5.5, is now broken because it is not possible to determine from the result set whether it is this special "output parameters" result set, from the metadata, until the whole result set is read. Because somebody wanted optimize out the 9 bytes in the result set metadata.

CLIENT_DEPRECATE_EOF decided to encode key-value in its own, unique way, dismissing the 20 years old notion of "result set" in the protocol and misused OK packet for this purpose. It was definitely a  low quality design work by newly hired  MySQL trainee by Oracle. Someone needs to stop these people from screwing the protocol.

bibstha:
@vaintroub that's interesting insight regarding compatibility.

Given that is how the protocol works now, and most other connectors have it implemented, and without `CLIENT_DEPRECATE_EOF`, the connector cannot access session tracking information, do you see any other way forward?

vaintroub:
I do not know about the good way forward. Luckily in this situation I'm almost a bystander/observer.

But the snarky remark I told, is the same I told maybe back in 2014,  when MySQL trainee developers successfully conquered their first challenge of the protocol butchering.
 
What now? 

We're debating to implement a MariaDB protocol extension, for our server, and our clients which would workaround what this CLIENT_DEPRECATE_EOF broke.  We need extension for https://jira.mariadb.org/browse/MDEV-19237 anyway, so maybe we can include the status flags that newbie developers throw away in intermediate EOF (2 of the whooping 9 bytes they spared per result set) 

I guess most of the client  APIs are doing OK without that extra trailing stuff at the end of OK packet. They did it for decades, and they can work with servers not supporting it. There is no notion of "client change state notification" in any of standard database APIs I know of. This stuff is mostly special needs for proxies, I think

vaintroub:
@bibstha , now what do *you* need this capability for, exactly? Can you replace it with something more well-formed, e.g a SELECT for a variable you're interested in, after the "COMMIT". In case you're concerned about network roundtrips, semicolon batching is your friend

bibstha:
These projects:

* [mariadb-corporation/mariadb-connector-j](https://github.com/mariadb-corporation/mariadb-connector-j)
* [mariadb-corporation/mariadb-connector-cpp](https://github.com/mariadb-corporation/mariadb-connector-cpp)

both already support the CLIENT_DEPRECATE_EOF. Wouldn't it simplify therefore to also have support in the c-connector? Maybe a flag to enable disable the support on client side?

The primary use-case for us to have this support is to be able to read session-tracking data. We are very much interested to gather query-metadata (specially performance) and return it in-line (realtime) so we can monitor / build systems that analyze queries behavior. The best way to do this for now is to have it return in the OK_packet as a session tracking variable.

We haven't considered batching queries yet, that's something we can look into. But the larger issue is still how you guys decide to move forward with the protocol support.

> There is no notion of "client change state notification" in any of standard database APIs I know of

> We're debating to implement a MariaDB protocol extension

IMO having a way to send back metadata of a query is very very useful. If not session_tracking, while you guys discuss protocol extension, please consider someway of sending back the metadata.

9EOR9:
Closing PR - we will not implement DEPRECATE_EOF due to broken design (see CONC-461)

PR #134, opened 2020-04-04
mariadb_config: solaris fix - types for options

grooverdan:
Not 100% sure I like OPT_STRING_TYPE as a name but it works.

I submit this under the MCA.


grooverdan:
thanks

PR #135, opened 2020-05-19
Fix MinGW build by adding 2 manual #defines

poelmanc:
Building with MinGW reports SP_PROT_TLS1_1_CLIENT and SP_PROT_TLS1_2_CLIENT as undefined. This fix follows the same approach as mariadb-connector-c/plugins/auth/sspi_errmsg.c - defining them in the .c file if and only if they're undefined.

### DIFF
--- a/libmariadb/secure/ma_schannel.c
+++ b/libmariadb/secure/ma_schannel.c
@@ -23,6 +23,15 @@
 #include "schannel_certs.h"
 #include <assert.h>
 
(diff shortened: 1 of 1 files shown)


9EOR9:
Closing PR since the definitions were already added into mingw headers (schannel.h)

poelmanc:
Excellent, I'm kind of a MinGW newbie and in retrospect MinGW's schannel.h was certainly where it really belonged.

PR #136, opened 2020-05-19
Fix MinGW build, adding #defines and a struct

poelmanc:
Building with MinGW reports SP_PROT_TLS1_0_CLIENT, SP_PROT_TLS1_1_CLIENT, SP_PROT_TLS1_2_CLIENT, SECPKG_ATTR_CIPHER_INFO, SECPKGCONTEXT_CIPHERINFO_V1, and struct _SecPkgContext_CipherInfo as undefined. This fix follows the same approach as mariadb-connector-c/plugins/auth/sspi_errmsg.c - defining them in the .c file if and only if they're undefined - and adds the struct as defined by Microsoft if and only if SECPKGCONTEXT_CIPHERINFO_V1 is undefined.

I considered an alternate approach of adding these to a header file so they could be shared across files. But that creates potential include order issues - if that header file were included before including Microsoft header files that correctly define them, developers might get macro redefinition errors - so keeping this compatibility logic in the .c files that need it seemed safer.

### DIFF
--- a/libmariadb/secure/schannel.c
+++ b/libmariadb/secure/schannel.c
@@ -32,6 +32,50 @@ char tls_library_version[] = "Schannel";
 #define PROT_TLS1_2 4
 #define PROT_TLS1_3 8
 
+/* These may not be defined e.g. on MINGW. */
+/* Values from https://docs.microsoft.com/en-us/windows/win32/api/schannel/ns-schannel-secpkgcontext_connectioninfo */
+#ifndef SP_PROT_TLS1_CLIENT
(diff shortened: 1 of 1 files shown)


9EOR9:
Why don't you add these definitions in MinGW project? That looks much simpler than changing every project which uses schannel.

Biswa96:
mingw-w64 is constantly updating. Please try with latest one. Those are added a while ago. You could use the mingw-w64 headers provided by msys2.

9EOR9:
Closing this PR since the missing definitions were already added in mingw headers.

PR #139, opened 2020-06-14
Add support for kFreeBSD builds

ottok:
Upstreamed from
https://salsa.debian.org/mariadb-team/mariadb-10.4/-/blob/master/debian/patches/libmariadb-kfreebsd.patch


9EOR9:
Hi Otto,

thanks for your contribution. Merged into 3.1

PR #143, opened 2020-08-04
Add Xpand to mysql_get_server_name

mariadb-michae2:
When connecting to the Xpand backend with the mariadb command-line client, the default prompt says "MySQL> ". This can be confusing, so usually we recommend that people change it with the prompt command. But it would be nice if the client library simply knew that it was Xpand and not MySQL (or MariaDB).

### DIFF
--- a/libmariadb/mariadb_lib.c
+++ b/libmariadb/mariadb_lib.c
@@ -3865,13 +3865,19 @@ my_bool STDCALL mariadb_connection(MYSQL *mysql)
           strstr(mysql->server_version, "-maria-"));
 }
 
+static my_bool xpand_connection(MYSQL *mysql)
+{
+  return (strstr(mysql->server_version, "-xpand-") != NULL);
+}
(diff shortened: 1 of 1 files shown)


9EOR9:
Hi,

can you provide an example for the server_version string sent in the server_hello packet?

mariadb-michae2:
Yes, no problem. In my current development build it is: `5.0.45-xpand-glassbutte-16592-438cd7fec881e6ea`
A release build of Xpand 9.3 Beta 9 looks like: `5.0.45-xpand-9.3_beta_9`

9EOR9:
I think this will cause a lot of problems:

- clients will interpret version number as 5.0.45 (C/C checks for `5.5.5-` prefix) and might not support features which were added in 5.1 and above.
- clients will not recognize anymore that they connect against a MariaDB server and will not support certain features. (see api function mariadb_connection).

PR #147, opened 2020-08-29
[CONC-491] Fix compilation with GNU iconv on OpenBSD

brad0:
This fixes building with GNU iconv on OpenBSD and other OS's using GNU iconv as a non native implementation.


brad0:
Looks like some of the CI tests were broken and have since been fixed. Any way to get this run through a CI test again?

9EOR9:
With rev. 7052619cc4b447da54a0b1458e096bb8dc4b3c77 Connector/C is built by default without iconv. Iconv support will be removed in next major version of MariaDB Connector/C.

PR #151, opened 2020-10-07
Fix race condition on global max_allowed_packet and net_buffer_length

alesapin:
Hello! In [ClickHouse DBMS](https://github.com/ClickHouse/ClickHouse) we use `mariadb-connector-c` for interactions with MySQL. We have a lot of tests for this functionality which runs with different [sanitizers](https://github.com/google/sanitizers). Recently thread sanitizer detected race condition in `mariadb-connector-c`: [link](https://gist.github.com/alesapin/1fa65d66dd24be123abba86c25410be5). Seems like this race caused by changes in global variables from different threads which cannot be controlled by the library user code.

Fix in this PR is the most straightforward way to avoid race condition. 

Thanks for the great library.

### DIFF
--- a/include/mariadb_com.h
+++ b/include/mariadb_com.h
@@ -382,8 +382,17 @@ enum enum_field_types { MYSQL_TYPE_DECIMAL, MYSQL_TYPE_TINY,
 #define FIELD_TYPE_GEOMETRY MYSQL_TYPE_GEOMETRY
 #define FIELD_TYPE_BIT MYSQL_TYPE_BIT
 
-extern unsigned long max_allowed_packet;
-extern unsigned long net_buffer_length;
+#if defined(__cplusplus) && defined(__GNUC__) && !defined(__clang__)
+extern "C++" {
+    #include <atomic>
+}
+
+#define _Atomic(T) std::atomic<T>
+
+#endif
+
+extern _Atomic(unsigned long) max_allowed_packet;
+extern _Atomic(unsigned long) net_buffer_length;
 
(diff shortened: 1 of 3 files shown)


9EOR9:
Thanks for your contribution, I filed the bug in our ticket system ([CONC-507](https://jira.mariadb.org/browse/CONC-507)).

I think just removing the setting of max_allowed_packet in ma_net_init() should avoid any race condition. Using _Atomic() for variable declaration might be another option, however this is not portable since MSVC still doesn't support all C11 extensions.

Could you please check, if the latest commit rev. ca4f043cfe8ea84fd9d215de8f0a4b75db18fa62  would fix your problem?

alesapin:
> Thanks for your contribution, I filed the bug in our ticket system ([CONC-507](https://jira.mariadb.org/browse/CONC-507)).
> 
> I think just removing the setting of max_allowed_packet in ma_net_init() should avoid any race condition. Using _Atomic() for variable declaration might be another option, however this is not portable since MSVC still doesn't support all C11 extensions.
> 
> Could you please check, if the latest commit rev. [ca4f043](https://github.com/mariadb-corporation/mariadb-connector-c/commit/ca4f043cfe8ea84fd9d215de8f0a4b75db18fa62) would fix your problem?

It was my initial solution for the problem, but I was not sure, that we really don't need to change `max_allowed_packet` here :)

Thank you, I think this PR can be closed.

9EOR9:
Closed on request.

PR #153, opened 2021-01-18
test fixes

Antchel:
I found some mistakes and error in tests. May be it will be useful.


9EOR9:
Thanks for the contribution - we will merge it after C/C 3.1.12 (and Server 10.5.9) release

PR #158, opened 2021-03-02
Fix flow control statement

Whissi:
cmake-3.20 has added checks for invalid flow control statements causing mariadb-connector-c to fail in configure phase with
```
CMake Error at cmake/ConnectorName.cmake:30 (ENDMACRO):
  Flow control statements are not properly nested.
Call Stack (most recent call first):
  CMakeLists.txt:423 (INCLUDE)
```


an3l:
This is already pushed with https://github.com/mariadb-corporation/mariadb-connector-c/commit/242cab8c

PR #159, opened 2021-03-05
IF()/ELSE() needs ENDIF(), not END()

poelmanc:
Prior CMake versions evidently accepted END() instead of ENDIF(), perhaps with a warning? We've been using this code for quite a while and it has worked. However, the latest CMake 3.20 release candidate treats END() as an error.


9EOR9:
Thanks for your contribution - added to 3.1

PR #163, opened 2021-03-25
MDEV-8334: Rename utf8 to utf8mb3

rucha174:
This patch is made as a part of MDEV-8334 to fix failing test in unit and
main test suite so that utf8mb3 characterset is recognized. Failing tests:
main.mysql_client_test
main.mysql_client_test_comp
unit.conc_basic-t
unit.conc_charset
unit.conc_connection

### DIFF
--- a/libmariadb/mariadb_lib.c
+++ b/libmariadb/mariadb_lib.c
@@ -3796,6 +3796,8 @@ int STDCALL mysql_set_character_set(MYSQL *mysql, const char *csname)
 
   if (!csname)
     goto error;
+  if (!strcasecmp("utf8",csname))
+    csname= "utf8mb4";
 
   if ((cs= mysql_find_charset_name(csname)))
   {
--- a/unittest/libmariadb/basic-t.c
+++ b/unittest/libmariadb/basic-t.c
@@ -121,7 +121,7 @@ static int test_conc71(MYSQL *my)
   mysql= mysql_init(NULL);
 
 
-  mysql_options(mysql, MYSQL_SET_CHARSET_NAME, "utf8");
+  mysql_options(mysql, MYSQL_SET_CHARSET_NAME, "utf8mb3");
   mysql_options(mysql, MYSQL_OPT_COMPRESS, 0);
   mysql_options(mysql, MYSQL_INIT_COMMAND, "/*!40101 SET SQL_MODE='' */");
   mysql_options(mysql, MYSQL_INIT_COMMAND, "/*!40101 set @@session.wait_timeout=28800 */");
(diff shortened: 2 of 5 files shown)


libmariadb/mariadb_lib.c
```diff
@@ -3796,6 +3796,8 @@ int STDCALL mysql_set_character_set(MYSQL *mysql, const char *csname)
 
   if (!csname)
     goto error;
+  if (!strcasecmp("utf8",csname))
+    csname= "utf8mb4";
```
9EOR9, line 3800: I wonder why In ma_charset you set character to utf8mb3 (if is "utf8"), but here you set it to uf8mb4?!

unittest/libmariadb/basic-t.c
```diff
@@ -310,7 +310,7 @@ static int use_utf8(MYSQL *my)
 
   while ((row= mysql_fetch_row(res)) != NULL)
   {
-    FAIL_IF(strcmp(row[0], "utf8"), "wrong character set");
+    FAIL_IF(strcmp(row[0], "utf8mb3"), "wrong character set");
```
9EOR9, line 313: Will server versions < 10.6 return utf8mb3 or utf8? If latter one, this change will break travis tests (testing 10.2 - 10.5). Maybe better to check both: `FAIL_IF(strcmp(row, "utf8") && strcmp(row[0], "utf8mb3"), ...)`

9EOR9 (commented):
Please make sure that travis tests will pass (except 10.6, unless MDEV-8334 was pushed).

unittest/libmariadb/basic-t.c
9EOR9: after reading MDEV-8334 more careful it seems that 10.7 or later will use utf8mb4 instead of utf8mb3.. In this case I think it is better, not to change existing tests (setting utf8mb3 instead of utf8) but to check the results depending on the server version:

```
const char *get_utf8_name(MYSQL *mysql)
{
  unsigned long server_version= mysql_get_server_version(mysql);

  if (server_version >= 100700)
    return "utf8mb4";
  if (server_version >= 100600)
    return "utf8mb3";
  return "utf8";
}
```
and for checking the character set

`FAIL_IF(strcmp(row[0], get_utf8_name(mysql), "Error message");`

libmariadb/mariadb_lib.c
rucha174: Typo, sorry. Shall fix.

unittest/libmariadb/basic-t.c
rucha174: I hear we use 3.2 and above for server version 10.6 and up(and 3.1 for < 10.6). and this patch is to be added on 3.2. I am sorry, I don't quite understand why it would affect version <  10.6 if we are not making changes on 3.1

PR #172, opened 2021-06-06
CONC-553: Fix SIGBUS when calling mysql_real_connect_start on ARM64 macOS

kotori2:
From [this stack overflow question](https://stackoverflow.com/questions/23852522/swapcontext-segfaults-when-swapping-to-a-ucontext-ts-uc-link)
> Evidently, that macro changes the layout of the relevant structures to match what those functions expect and require.

So I moved the `_XOPEN_SOURCE` macro to `CMakefile.txt`.


libmariadb/CMakeLists.txt
```diff
@@ -4,6 +4,9 @@ INCLUDE_DIRECTORIES(${CC_SOURCE_DIR}/include
 ADD_DEFINITIONS(-D HAVE_COMPRESS)
 ADD_DEFINITIONS(-D LIBMARIADB)
 ADD_DEFINITIONS(-D THREAD)
+IF(APPLE)
+  ADD_DEFINITIONS(-D _XOPEN_SOURCE)
```
vuvova, line 8: shouldn't it be `_XOPEN_SOURCE=600` ?

kotori2: I have no idea why it was `600` as I checked `ucontext.h`: 
```
#ifdef _XOPEN_SOURCE
...
#else /* !_XOPEN_SOURCE */
#error The deprecated ucontext routines require _XOPEN_SOURCE to be defined
#endif /* _XOPEN_SOURCE */
```

Also it seems it was producing compilation errors if I set `_XOPEN_SOURCE` to `600`. Setting `-D _XOPEN_SOURCE` just works fine.

vuvova: ok, fair enough

kotori2:
CI seems passing now.

0rangeFox:
@vuvova This is really necessary for M1 working when creating MySQL Client through code, and I confirm it's working perfectly with this simple changes.

9EOR9:
Manually merged into 3.1 branch

PR #179, opened 2021-08-12
test openssl on windows

grooverdan:
just testing


vaintroub:
Currently, the hashing and encryption stuff is always Windows, independent from OPENSSL, which is used for SSL, if configured.
Why would you want to change that?

grooverdan:
Trying to help out the R people who have trouble building a static connector-c on Windows against SCHANNEL which seems to relate to plugins ed25519 and caching_sha2_password won't compile.

https://github.com/r-dbi/RMariaDB/issues/134#issuecomment-769819845
https://github.com/r-windows/rtools-packages/blob/master/mingw-w64-libmariadbclient/PKGBUILD#L43

vaintroub:
> Trying to help out the R people who have trouble building a static connector-c on Windows against SCHANNEL which seems to relate to plugins ed25519 and caching_sha2_password won't compile.
> 
> [r-dbi/RMariaDB#134 (comment)](https://github.com/r-dbi/RMariaDB/issues/134#issuecomment-769819845)
> https://github.com/r-windows/rtools-packages/blob/master/mingw-w64-libmariadbclient/PKGBUILD#L43

Their comment is wrong, BCrypt requires Windows Vista, which is by now 15 years old. I do not think Windows XP support is still something worth pursuing. https://docs.microsoft.com/en-us/windows/win32/api/bcrypt/nf-bcrypt-bcryptcreatehash

Perhaps, they'd also need mingw that is capable of using Vista APIs.

PR #182, opened 2021-09-29
Increase HOSTNAME_LENGTH limit to 255 bytes

XYHong18:
The HOSTNAME_LENGTH was increased to 255 bytes in MariaDB 10.6 in
commit https://github.com/MariaDB/server/commit/a3099a3b.

To make it consistent, update HOSTNAME_LENGTH in mariadb-connector-c
to 255 bytes as well.

All new code of the whole pull request, including one or several files that are
either new files or modified ones, are contributed under the BSD-new license. I
am contributing on behalf of my employer Amazon Web Services, Inc.


SergMariaDB (approved):
BTW, TBD, its include guard is the same as server/include/mysql_com.h has.

PR #183, opened 2021-09-29
Xcode compatibility update

ohhmm:
Built with latest Xcode for Apple Silicon


vaintroub:
Hmm, why would you want both -Wno-deprecated-declarations and pragmas, which say the same thing? Also, `#ifdef __APPLE__  ` does not necessarily mean clang, there is gcc on macOS we well. if you like pragmas more, it better be ` #pragma GCC ` , it is applicable to both clang and GCC, as clang impersonates GCC. But  with compiler flag, it is much less code.

SergMariaDB (approved):
TODO : implement yet another ARM64-portable way for future if Apple removes ucontext.h

PR #185, opened 2021-10-18
make libmariadb able to be builded as static library

samuaz:
I'm working on a cpp backend that runs as a single binary in a minimal docker image, so we need that binary can run without any system deps. so this PR make able to build libmariadb as static library.


9EOR9:
Why can't you use mariadbclient static library provided by Connector/C ?

9EOR9:
Closing due to missing response

samuaz:
@9EOR9 sorry for the delay.

Because it builds as shared lib, so then when I build my binary as static with this as dependency complains about undefined from this.

This PR just add the ability to build as static lib.

PR #187, opened 2021-11-19
Add support for fake ROTATE_EVENT

kou:
https://mariadb.com/kb/en/fake-rotate_event/

### DIFF
--- a/libmariadb/mariadb_rpl.c
+++ b/libmariadb/mariadb_rpl.c
@@ -309,7 +309,31 @@ MARIADB_RPL_EVENT * STDCALL mariadb_rpl_fetch(MARIADB_RPL *rpl, MARIADB_RPL_EVEN
     case ROTATE_EVENT:
       rpl_event->event.rotate.position= uint8korr(ev);
       ev+= 8;
-      len= rpl_event->event_length - rpl->fd_header_len - 8;
+      if (rpl_event->timestamp == 0)
+      {
+        /* Fake ROTATE_EVENT: https://mariadb.com/kb/en/fake-rotate_event/
+           This is sent before FORMAT_DESCRIPTION. So we can't use
+           rpl->use_checksum and rpl->fd_header_len here.
+         */
+        const uint8_t header_size = 19;
+        len= rpl_event->event_length - header_size - 8;
+        /*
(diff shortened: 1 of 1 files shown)


libmariadb/mariadb_rpl.c
```diff
@@ -309,7 +309,31 @@ MARIADB_RPL_EVENT * STDCALL mariadb_rpl_fetch(MARIADB_RPL *rpl, MARIADB_RPL_EVEN
...
+      if (rpl_event->timestamp == 0)
+      {
+        /* Fake ROTATE_EVENT: https://mariadb.com/kb/en/fake-rotate_event/
+           This is sent before FORMAT_DESCRIPTION. So we can't use
+           rpl->use_checksum and rpl->fd_header_len here.
+         */
+        const uint8_t header_size = 19;
+        len= rpl_event->event_length - header_size - 8;
+        /*
+          How can we detect whether checksum is used or not?
```
andrelkin, line 321: I looks that we can, and I believe you could not have missed `MARIADB_RPL::is_checksum`.
The member may get evaluated before the very first FD is fetched a way the slave server currently does.
That is to essentially query select @@global.master_binlog_checksum. If you have not read the following snippet, please do(, otherwise why not do it this way?):
```
/**
  Function for calling in mysql_binlog_send
  to get the value of @@binlog_checksum of the master at
  time of checksum-handshake.

  The value tells the master whether to compute or not, and the slave
  to verify or not the first artificial Rotate event's checksum.

  @param[in]    thd  THD to access a user variable

  @return       value of @@binlog_checksum alg according to
                @c enum enum_binlog_checksum_alg
*/

static enum enum_binlog_checksum_alg get_binlog_checksum_value_at_connect(THD * thd)
```

9EOR9: @andrelkin  Can't we rely on checksum algorithm which was sent in [Format description event](https://mariadb.com/kb/en/format_description_event) instead of checking global variable

9EOR9:
added and fixed in rev. abddf0b0af6c20ca89e9085c1dc3bea9d2ac580f

kou:
Thanks!
(Sorry for not following up this...)

PR #188, opened 2021-11-19
Add support for ROWS_EVENT V2

kou:
ROWS_EVENT V2 has the extra-data field after the flags field in
header.

See also: https://dev.mysql.com/doc/internals/en/rows-event.html


kou:
@9EOR9 Could you review this? Please let me know if you have any missing information.

kou:
@9EOR9 Could you also review this?

PR #191, opened 2022-02-26
Stop constant relink after cmake

poelmanc:
The prior CMake logic wrote the .def file whenever CMake was run. This forced the next build to relink the library, even when nothing changed, which forced all downstream libraries to relink.

This change simplifies the CMake logic and only writes the .def file when its contents change.

### DIFF
--- a/cmake/export.cmake
+++ b/cmake/export.cmake
@@ -5,26 +5,28 @@
 #  BSD license.
 #  For details see the COPYING-CMAKE-SCRIPTS file.
 #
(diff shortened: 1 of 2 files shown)


poelmanc:
Sorry, I based this patch on 'master' which I now see is dead. Does general new development like these cmake fixes belong on '3.3' perhaps or elsewhere?

9EOR9:
default branch is 3.3 - master is not in use anymore and will be removed.

PR #194, opened 2022-03-31
Copy proxy header for later use

markus456:
If the memory passed as the argument was freed after the mysql_optionsv
call, the values would point to freed memory. Keeping a copy of the
buffer allows the original buffer to be freed right after the
mysql_optionsv call.

### DIFF
--- a/libmariadb/mariadb_lib.c
+++ b/libmariadb/mariadb_lib.c
@@ -1990,6 +1990,7 @@ static void mysql_close_options(MYSQL *mysql)
     free(mysql->options.extension->tls_version);
     free(mysql->options.extension->url);
     free(mysql->options.extension->connection_handler);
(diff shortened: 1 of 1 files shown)


markus456:
Given that the connector supports both versions 1 and 2 of the proxy protocol, it must also be able to deal with binary data. I mistakenly assumed that it would only support the version 1 header.

markus456:
This should also solve the (theoretical, at this point) problem of using `MYSQL_OPT_RECONNECT` with proxy protocol. I'd say it's less likely that the memory used to store the string is around after the `mysql_real_connect` call and should lead to a similar use-after-free problem.

9EOR9:
Thanks for your PR - I've modified and pushed your patch (added new macro OPT_SET_EXTENDED_VALUE_BIN).

PR #197, opened 2022-05-24
make libmariadb able to be buidled as static library

samuaz:
Hi guys,

im working on app builded statically to run on a docker alpine without need of any system shared library. 

but in this cmake the library is always builded as shared.

This PR just add the ability to build as static lib or shared as you need.


libmariadb/CMakeLists.txt
```diff
@@ -4,6 +4,7 @@ INCLUDE_DIRECTORIES(${CC_SOURCE_DIR}/include
 ADD_DEFINITIONS(-D HAVE_COMPRESS)
 ADD_DEFINITIONS(-D LIBMARIADB)
 ADD_DEFINITIONS(-D THREAD)
+ADD_OPTION(BUILD_SHARED_LIBS "Build as a shared library" ON)
```
9EOR9, line 7: I don't think thats a good idea. If you don't need a shared library, just don't use it (or remove) it. The overhead for building shared/static  libraries is minimal, since they are created from object library.

samuaz: @9EOR9 thanks for the reply.

Can you give me more details please? You mean to remove the default ON for this option?

or you mean that this change is not needed to achieve what im looking?

my context its:

- im building a backend application that use mariadb connector, 
- that app is build as static 
- my binary need to be able to work alone without need of the shared libs
- the binary lives inside alpine docker image that don't have any other deps than my binary, that means the mariadb lib is not in the system as a shared lib
- that alpine docker image is used on gcloud run

If mariadb is not build as static and linked to my binary, when I execute my binary, I have the Undefined Reference error.

Having mariadb build statically and linking the .a library with my binary, not an issue.

9EOR9:
no feedback -> closed.

libmariadb/CMakeLists.txt
FooBarrior: >that means the mariadb lib is not in the system as a shared lib

actually that doesn't. You shouldn't necessarily put a shared lib in a system folder. All you need is specify LD_LIBRARY_PATH for the app to run (or ldconfig it). Anyway, if you don't need a shared object, you can just rm it with our packaging script -- I guess that was @9EOR9's point.

sazconaitbit: Thanks for your response.  But I need to put that shared library on the docker alpine also right? I mean the libmaria…so 

I’m looking for a way to be able to build my binary statically with libmariadb and don’t need to have the shared library on the docker alpine image. 

The way that I found was build libmaria statically and then link with my binary, by apply the changes from this PR.

so my intention is just to have a way to build Mariadb as static library

lawrinn: Static libmariadb is named mariadbclient, and it's always built.
ADD_LIBRARY(mariadbclient STATIC ${MARIADB_OBJECTS} ${EMPTY_FILE})
You don't need it additionally built under another name

melroy89: I would you loved this `BUILD_SHARED_LIBS` option xD.. so I could turn it off !

Currently I need to set `INSTALL_PLUGINDIR` as well.. which I do not want. or need!

Also do not use `add_library()` use `target_link_libraries()`

melroy89:
> no feedback -> closed.

@9EOR9  Could we still introduce `BUILD_SHARED_LIBS` option..? So I can turn it `OFF`?  

I need workaround like setting `set(INSTALL_PLUGINDIR {CMAKE_BINARY_DIR}/lib)` as well currently.

I just want to use the static `libmariadbclient.a` basically..  I would be cool if I could turn off the shared stuff. Especially if I want to build my project across multiple platforms and cross compiling to Mac and Windows etc.

```cmake
include(FetchContent)
FetchContent_Declare(
  mariadb-connector-c
  GIT_REPOSITORY https://github.com/mariadb-corporation/mariadb-connector-c.git
  GIT_TAG        v3.4.8
)
# Dummy install dir so mariadbc doesn't get mad
set(INSTALL_PLUGINDIR {CMAKE_BINARY_DIR}/lib)

set(WITH_UNIT_TESTS OFF CACHE BOOL "" FORCE)
set(WITH_CURL OFF CACHE BOOL "" FORCE)
set(WITH_DYNCOL OFF CACHE BOOL "" FORCE)
set(MARIADB_LINK_DYNAMIC OFF CACHE BOOL "" FORCE)

if (CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
    # mariadb-c doesn't set this if its a subproject for some reason, even though they are needed for it to build.
	ADD_DEFINITIONS(-DWIN32_LEAN_AND_MEAN -DNOGDI)
endif()

FetchContent_MakeAvailable(mariadb-connector-c)
```

PR #211, opened 2023-01-06
Fixed unexpected blocking in async codepaths w/SSL and other issues

jtrberg:
- fixed https://jira.mariadb.org/browse/CONC-594
- fixed disabled 0s timeouts not handled correctly
- fixed pvio double free if async connect is aborted early via forced timeout


jtrberg:
I ran the unit test test_conc622 that reported a failure above locally in my branch:

```
21: # Expected error when connection to host '0.0.0.0'
21: not ok 5 - test_conc622
21: # close default
21: # Failed 1 tests!
21/21 Test #21: async ............................***Failed    0.12 sec
```

 but cannot repro the issue.

PR #217, opened 2023-02-07
WIP - Connection plugin to support multi-source load balancing

jonahharris:
Based on the master/slave connection plugin, add support for random-based load balancing for use against a multi-source cluster (in this case Xpand.)

`multisource://host1:port1,...,hostN:portN`

### DIFF
--- a/plugins/connection/multimaster.c
+++ b/plugins/connection/multimaster.c
@@ -1,0 +1,224 @@
+/************************************************************************************
+    Copyright (C) 2015-2023 MariaDB Corporation AB
+   
+   This library is free software; you can redistribute it and/or
+   modify it under the terms of the GNU Library General Public
+   License as published by the Free Software Foundation; either
+   version 2 of the License, or (at your option) any later version.
+   
+   This library is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+   Library General Public License for more details.
+   
+   You should have received a copy of the GNU Library General Public
+   License along with this library; if not see <http://www.gnu.org/licenses>
+   or write to the Free Software Foundation, Inc., 
+   51 Franklin St., Fifth Floor, Boston, MA 02110, USA
+
+   Part of this code includes code from the PHP project which
+   is freely available from http://www.php.net
+*************************************************************************************/
+
+/* MariaDB Connection plugin for load balancing  */
+
+#include <ma_global.h>
+#include <ma_sys.h>
+#include <errmsg.h>
+#include <mysql.h>
+#include <mysql/client_plugin.h>
+#include <string.h>
+#include <ma_string.h>
+#include <ma_common.h>
+
+#ifndef WIN32
+#include <sys/time.h>
+#endif
+
+/* function prototypes */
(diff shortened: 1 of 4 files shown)


plugins/connection/multimaster.c
```diff
@@ -0,0 +1,224 @@
...
+#include <string.h>
+#include <ma_string.h>
+#include <ma_common.h>
+
+#ifndef WIN32
+#include <sys/time.h>
+#endif
+
+/* function prototypes */
+MYSQL *multimaster_connect(MYSQL *mysql, const char *host, const char *user, const char *passwd,
```
9EOR9: See https://jira.mariadb.org/browse/MDEV-18777. We use nowadays source and replica instead of master and slave. According to https://mariadb.com/kb/en/multi-source-replication/ the correct name would be multisource.

plugins/connection/multimaster.c
```diff
@@ -0,0 +1,224 @@
...
+  {
+    *p= '\0';
+    p++;
+    data->port= atoi(p);
+  }
+
+  return 0;
+}
+
+MYSQL *multimaster_connect(MYSQL *mysql, const char *host, const char *user, const char *passwd,
```
9EOR9: If I understand this code correctly, this is already implemented in mysql_real_connect, where you can specify a list of different hosts in hostname. This is used for simple failover: If we can't connect to server1, we will try to connect to server2, ....
Maybe we should save this information, so redundant parsing will not be needed. host_list could be retrieved then via mariadb_get_optionv().

jonahharris: Agreed - the main issue is we need to randomly pick an Xpand server from the cluster given a singular connect string. The randomization approach we use in our code is similar to what you have for the source/replica connection plug-in secondary selection, so it seemed this simpler plugin was the best approach. There are definitely areas for optimization.

Interested in your overall thoughts - talking to Max I think the goal is for the C connector to have general load-balancing support inherently similar to connector-j, which could be done internally in libmariadb's connection APIs or with a connector like this (renamed something like loadbalance or the like) and made static.

I haven't written a client-side plug-in before so I'm still looking through library in terms of the expectations as well as what needs to be maintained in the plug-in. At first glance, it doesn't seem we need to maintain our own copy of the pvio and such in this case, unlike your original one which had to deal with multiple open sockets.

plugins/connection/multimaster.c
jonahharris: Thanks, @9EOR9 - I saw that naming in the current version and maintained it even though, as you say, source/replica are generally used now.

jonahharris: Updated.

rusher:
Just a thought, but if doing like java, there is a few things to add : 
* when failing to connect, host/port combination will be in a static denied list for some time (60s by default, can be set by option)
* in place or using a random distribution, use a real round-robin distribution

jonahharris:
> Just a thought, but if doing like java, there is a few things to add :
> 
> * when failing to connect, host/port combination will be in a static denied list for some time (60s by default, can be set by option)

Agreed. That needs to be accounted-for and is not in the code at present. 

> * in place or using a random distribution, use a real round-robin distribution

While I agree round-robin would be a better overall approach, we can't really use it as the client using the connector library is multi-process; each long-lived process forks, gets its own copy of the identical connection string, and would connect to the Xpand servers in the same order, consistently overloading the first ones in the list.

PR #218, opened 2023-02-20
Fix compiler warnings

markus456:
Fixed the possible uninitialized use of the decimal types. This isn't a real problem but helps keep the compilers happy.


markus456:
The crypto stuff is no longer a problem. Changed the commit to fix the warnings.

markus456:
Rebasing on top of 3.3.

markus456:
Oh, it was already fixed.

PR #222, opened 2023-06-05
MDEV-30587: TLS/SSL error: unexpected eof while reading

grooverdan:
Since OpenSSL-3.0 there is this warning if the peer unexpected shutdown.

As MariaDB uses SSL_set_quiet_shutdown on the server side to perform an abrupt shutdown this leaves the client issuing warnings often on every TLS connection.

ref: https://github.com/openssl/openssl/issues/18866#issuecomment-1194219601


vuvova:
No, why would we want to hide a valid and perfectly sensible OpenSSL 3.x error? Let's not do that

PR #224, opened 2023-06-15
[MDEV-28634] Do not allow silent downgrade of connections where SSL is requested

dlenski:
The 2015 commit
https://github.com/mariadb-corporation/mariadb-connector-c/commit/23895fbd4#diff-4339ae6506ef1fb201f6f836085257e72c191d2b4498df507d499fc30d891005
was described as "Fixed gnutls support", which is an extremely incomplete
description of what it actually does.

In that commit, the Connector/C client authentication plugin was modified to
abort following the initial server greeting packet if the client has
requested a secure transport (`mariadb --ssl`), but the server does not
advertise support for SSL/TLS.

However, there's a crucial caveat here:

If the client has requested secure transport (`mariadb --ssl`) but has not
requested verification of the server's TLS certificate
(`mariadb --ssl --ssl-verify-server-cert`), then the client will silently
accept an unencrypted connection, without even printing a diagnostic message.

Thus, any such client is susceptible to a trivial downgrade to an
unencrypted connection by an on-path attacker who simply flips the TLS/SSL
capability bit in the advertised server capabilities in the greeting packet.

The entire design of MariaDB's TLS support is severely flawed in terms of
user expectations surrounding secure-by-default connections: the `--ssl`
option SHOULD imply `--ssl-verify-server-cert` BY DEFAULT; if a client
actually wants a TLS connection that's susceptible to a trivial MITM'ed
by any pervasive or on-path attacker, that should be an exceptional case
(e.g. `--insecure-ssl-without-server-cert-verification`) rather than the
default.

Even without resolving the issue of no default verification of server
certificates, the issue of silent downgrade to unencrypted connections can
and should be resolved.

Additionally, the second commit moves the check for server TLS/SSL capability from `send_client_reply_packet` (in `my_auth.c`) to `mthd_my_real_connect` (in `mariadb_lib.c`). Two reasons for this:

1. Reduction of attack surface

   As soon as the client receives the server's capability flags, it knows
   whether the server supports TLS/SSL.

   If the server does not support TLS/SSL, but the client expects and
   requires it, the client should immediately abort at this point in order
   to truncate any code paths by which it could inadvertently continue to
   communicate without TLS/SSL.

2. Separation of concerns

   Whether or not the server supports TLS/SSL encryption at the transport
   layer (TLS stands for TRANSPORT-layer security) is a logically separate
   issue from what APPLICATION-layer authentication modes the client and
   server support or should use.

# Backwards compatibility

This is an INTENTIONAL backwards-incompatible
change to prevent clients from being silently downgraded to unencrypted
connections, when they've specified `--ssl` and thus clearly indicated that
they do not want unencrypted connections.

### DIFF
--- a/libmariadb/mariadb_lib.c
+++ b/libmariadb/mariadb_lib.c
@@ -1918,6 +1918,16 @@ restart:
     }
   }
 
+  /* We now know the server's capabilities. If the client wants TLS/SSL,
+   * but the server doesn't support it, we should immediately abort.
+   */
+  if (mysql->options.use_ssl && !(mysql->server_capabilities & CLIENT_SSL))
+  {
+    SET_CLIENT_ERROR(mysql, CR_SSL_CONNECTION_ERROR, SQLSTATE_UNKNOWN,
+                     "Client requires TLS/SSL, but the server does not support it");
+    goto error;
+  }
+
   /* Set character set */
   if (mysql->options.charset_name)
     mysql->charset= mysql_find_charset_name(mysql->options.charset_name);
--- a/plugins/auth/my_auth.c
+++ b/plugins/auth/my_auth.c
@@ -245,15 +245,6 @@ static int send_client_reply_packet(MCPVIO_EXT *mpvio,
     mysql->server_capabilities &= ~(CLIENT_SSL);
   }
 
-  /* if server doesn't support SSL, we need to return an error */
-  if (mysql->options.use_ssl && !(mysql->server_capabilities & CLIENT_SSL))
-  {
-    my_set_error(mysql, CR_SSL_CONNECTION_ERROR, SQLSTATE_UNKNOWN,
-                        ER(CR_SSL_CONNECTION_ERROR), 
-                        "SSL is required, but the server does not support it");
-    goto error;
-  }
-
   /* Remove options that server doesn't support */
   mysql->client_flag= mysql->client_flag &
                        (~(CLIENT_COMPRESS | CLIENT_ZSTD_COMPRESSION | CLIENT_SSL | CLIENT_PROTOCOL_41) 
--- a/plugins/auth/my_auth.c
+++ b/plugins/auth/my_auth.c
@@ -245,21 +245,14 @@ static int send_client_reply_packet(MCPVIO_EXT *mpvio,
     mysql->server_capabilities &= ~(CLIENT_SSL);
   }
 
-  /* if server doesn't support SSL and verification of server certificate
-     was set to mandatory, we need to return an error */
+  /* if server doesn't support SSL, we need to return an error */
   if (mysql->options.use_ssl && !(mysql->server_capabilities & CLIENT_SSL))
   {
-    if ((mysql->client_flag & CLIENT_SSL_VERIFY_SERVER_CERT) ||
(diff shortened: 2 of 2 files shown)


vuvova:
let me mention a related https://jira.mariadb.org/browse/MDEV-28634 here, for cross-reference purposes

vuvova:
fixed in [MDEV-28634](https://jira.mariadb.org/browse/MDEV-28634)

dlenski:
@vuvova wrote:

> let me mention a related https://jira.mariadb.org/browse/MDEV-28634 here, for cross-reference purposes

This PR _has mentioned_ that JIRA in its title, since it was first created.

> fixed in [MDEV-28634](https://jira.mariadb.org/browse/MDEV-28634)

Really…? Where?

Which commit(s) in this repository address this specific problem?

vuvova:
Commit https://github.com/mariadb-corporation/mariadb-connector-c/commit/8dffd56936df3d03eeccf47904773860a0cdeb57 enables MYSQL_OPT_SSL_VERIFY_SERVER_CERT by default. When it's enabled the server will refuse to downgrade to non-ssl

dlenski:
@vuvova wrote:

> Commit https://github.com/mariadb-corporation/mariadb-connector-c/commit/8dffd56936df3d03eeccf47904773860a0cdeb57 enables MYSQL_OPT_SSL_VERIFY_SERVER_CERT by default. When it's enabled the server will refuse to downgrade to non-ssl

While that does _appreciably improve the default TLS behavior_ of Connector/C (and that's good :raised_hands:)…

1. It still remains _possible_ to configure the Connector/C client _not_ to do certificate validation
2. And if a user does that, it ***still remains possible*** for that connection to be _silently downgraded from TLS to plaintext_:
    https://github.com/mariadb-corporation/mariadb-connector-c/blob/8dffd56936df3d03eeccf47904773860a0cdeb57/plugins/auth/my_auth.c#L281-L292

What possible argument could you have for allowing that foot-gun to continue existing in the codebase? Particularly given that _this PR already exists_ and fixes it, and additionally limits the possible codepaths where cert validation could be bypassed, in https://github.com/mariadb-corporation/mariadb-connector-c/pull/224/commits/7d0c7035e64d364e392c5c5867662ead418ed7e2?

vuvova:
My argument always was that we aren't SSL police and aren't dictating the user what they can and cannot do. By default the client won't disable SSL, but if the user explicitly wants to shoot themselves in the foot — we aren't going to stop them. After all, even if the new client won't allow it, the user can always use an older client that still has supports silent downgrade.

dlenski:
> aren't dictating the user what they can and cannot do. 

>  if the user explicitly wants to shoot themselves in the foot — we aren't going to stop them.

When you maintain this attitude over many years and across widely different parts of the product, you end up with a product where there is a huge amount of confusion over the safe, correct, best practices for configuring and using it correctly.

I've made this point in numerous public and private communications with you, @vuvova, and in [my October presentation at the 2023 Unconference](https://mariadb.org/security-and-usability-of-tls-in-mariadb/).

> After all, even if the new client won't allow it, the user can always use an older client that still has supports silent downgrade.

You are using the existence of problems in older clients as part of a justification for _not fixing_ those problems in newer clients.

Why ever fix _anything at all_, then? :man_shrugging:

PR #225, opened 2023-06-24
Add pipelined execution for the binary protocol

markus456:
The mysql_send_query() function allows text protocol queries to be executed in a batched manner, also known as pipelined execution of the queries. The batched execution for the text protocol is done by sending the queries before reading their results. This allows the sending of the query and reading of the result to be detached which eliminates the latency that would otherwise exist for normal serialized execution.

The same can be done for the binary protocol by splitting the execution of mysql_stmt_execute() into separate parts. This commit adds the extensions mariadb_stmt_send_execute() and
mariadb_stmt_read_execute_result(). They behave the same way a normal mysql_stmt_execute() would: the query parameter binds are bound before the query is sent and the result binds are bound once the results are read.

With pipelined execution supported for both text and binary protocols, bulk insertions can be done more efficiently on servers that do not implement the MariaDB bulk insertion API.

I tested this with the `10.11.3` docker image and all tests pass.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -135,7 +135,7 @@ SET(CMAKE_CXX_FLAGS_RELWITHDEBINFO "${CMAKE_CXX_FLAGS_RELWITHDEBINFO} -DDBUG_OFF
 
 IF(CMAKE_COMPILER_IS_GNUCC)
   INCLUDE(CheckCCompilerFlag)
-  SET(GCC_FLAGS -Wunused -Wlogical-op -Wno-uninitialized  -Wall -Wextra -Wformat-security -Wno-init-self -Wwrite-strings -Wshift-count-overflow -Wdeclaration-after-statement -Wno-undef -Wno-unknown-pragmas)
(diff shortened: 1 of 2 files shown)


9EOR9 (commented):
Why don't use mysql_optionsv(mysql, MARIADB_OPT_SKIP_READ_RESPONSE, &bool) instead? See also CONC-533 and https://github.com/mariadb-corporation/mariadb-connector-c/commit/e4e5b28452815b6b6b82f2b6b032a9cc9613230c

markus456:
I was not aware that this option existed and what it did. Based on what the wikis says about this (it's the only reference I could find), I don't know if it even is the same.
> MARIADB_OPT_SKIP_READ_RESPONSE
Don't read response packets in binary protocol. Added in version 3.1.13

Would it be possible to improve the documentation with a usage example on how to use it with binary protocol prepared statements?

markus456:
I modified the commit to only add the new `mariadb_` functions to the API and converted them to use the option internally. Tests seem to pass with this version and it's less code which is usually better.

markus456:
I closed this since it doesn't seem to be required. I still think the existing APIs need to be properly documented as they are currently unusable due to the lack of documentation.

PR #226, opened 2023-06-28
[MDEV-15935] Client-side implementation of connection redirection mechanism

dlenski:
This is the **client** counterpart of https://github.com/MariaDB/server/pull/2681, which provides an initial **server**-side mechanism for redirecting new client connections.

It modifies the Connector/C library, so that if the server responds to our initial connection with an error packet having
error number 4196 (`ER_SERVER_REDIRECT`), and an error string formatted as `|Human-readable message|host[:port]`, the client will redirect its connection accordingly.

This redirection is performed in `mthd_my_real_connect`, which appears to be the most appropriate location given that it already contains code for retrying/repeating a new server connection.

---

This PR also incorporates fixes for two client-security vulnerabilities:

- [MDEV-28634](https://jira.mariadb.org/browse/MDEV-28634) ("improper TLS→plaintext downgrade"), also available as a standalone PR in https://github.com/mariadb-corporation/mariadb-connector-c/pull/224
- [CONC-648](https://jira.mariadb.org/browse/CONC-648) ("client improperly trusts server error packets sent prior to TLS handshake"), also available as a standalone PR in https://github.com/mariadb-corporation/mariadb-connector-c/pull/223

Both of these vulnerabilities _should be fixed_ regardless of whether a connection redirection mechanism is implemented, but their implications are very severe if a connection redirection mechanism exists (→ trivial for MITM attackers to redirect to attacker-controlled servers).


include/mysql.h
```diff
@@ -227,6 +227,8 @@ extern const char *SQLSTATE_UNKNOWN;
     /* MariaDB specific */
     MYSQL_PROGRESS_CALLBACK=5999,
     MYSQL_OPT_NONBLOCK,
+    MYSQL_OPT_FOLLOW_SERVER_REDIRECTS,
```
9EOR9: Should be prefixed with MARIADB unless MySQL/Oracle has the same option.

include/mysql.h
```diff
@@ -337,12 +339,19 @@ struct st_mysql_options {
     char *bind_address;
     my_bool secure_auth;
     my_bool report_data_truncation; 
+    my_bool follow_server_redirects;
```
9EOR9, line 342: Please add new options in mysql->options.extension to keep MYSQL structure binary compatible.

include/mysql.h
dlenski: Okay, will do…

But why are the immediately preceding options _both_ prefixed with `MYSQL` _and_ marked as `/* MariaDB specific */`?

Those are contradictory, are they not?

include/mysql.h
dlenski: I can do that but, as I commented [below](https://github.com/mariadb-corporation/mariadb-connector-c/pull/226/files#diff-e9cecafb403a5b293342e5d847287079094ba615dfaac282f22c82f55176091aR351), I don't understand this requirement:

> … to keep MYSQL structure binary compatible.

The structure `st_mysql` cannot possibly be used as a serialization format for inter-process communication, let alone network-based communication, because it contains pointers. and nothing is done to ensure consistent endianness or alignment.

So why does it need to be kept binary-compatible? With what?

dlenski: Done in https://github.com/mariadb-corporation/mariadb-connector-c/pull/226/commits/54886ac29d045cb1021cf82e4e20947c59ed5f23 and https://github.com/MariaDB/server/pull/2681/commits/4ffff66e0f72fe2a4ceeca029df7b35b4de1c4ef.

dlenski:
In https://github.com/mariadb-corporation/mariadb-connector-c/pull/226/commits/54886ac29d045cb1021cf82e4e20947c59ed5f23?w=1 and https://github.com/MariaDB/server/pull/2681/commits/4ffff66e0f72fe2a4ceeca029df7b35b4de1c4ef, I made `follow-server-redirection` a Connector/C option (enabled by default).

As discussed with @atcurtis on [the developers mailing list](https://lists.mariadb.org/hyperkitty/list/developers@lists.mariadb.org/message/G2NSDYPWXYCXHZN42R7WLW35VDUMGXY3/), I'll further…

- [ ] Update this PR so that `follow-server-redirection` is enabled _by default_ **if and only if** either `ssl-verify-server-cert` or `ssl-fp-list` is enabled.

vuvova:
@dlenski, I'm a bit confused. I thought it was a PoC patch to show how to do forced redirection with an error packet. And now you seems to work on making it the production quality. This is, of course, fine. But there's yet no decision, as far as I know, to use the error message approach. It'd be kind of a waste if a perfectly good patch would be rejected because it implements a wrong approach

vuvova:
We went on with the sysvar tracker approach as more flexible for a DBA to use.

It's unrelated to the forced/optional discussion, because a server can disconnect after sending the redirection info to the client either way, whether it sends an error message or a sysvar tracking packet.

PR #228, opened 2023-07-12
[CONC-656] Do not send CLIENT_CAN_VERIFY_SERVER_CERT bit on the wire

dlenski:
Sending this bit effectively tells MITM attackers whether the client will be able to *detect* a MITM attack using 2-sided TLS (e.g.  what `mitmproxy` does).

This vulnerability is a follow-on consequence of CONC-654, but can easily be fixed independently of that more complex issue.

We fix it by setting the `CLIENT_SSL_VERIFY_SERVER_CERT` bit to 0/*OFF*, rather than to 1/*ON*, in the pre-TLS SSLRequest packet; this makes fixed clients indistinguishable from older clients which foolishly vary this bit. Once many fixed clients are deployed, this will make it much harder for MITM attackers to use this bit as a reliable way to target vulnerable clients.

### DIFF
--- a/plugins/auth/my_auth.c
+++ b/plugins/auth/my_auth.c
@@ -297,10 +297,23 @@ static int send_client_reply_packet(MCPVIO_EXT *mpvio,
 
   if (mysql->client_flag & CLIENT_PROTOCOL_41)
   {
+    /* Whether or not the client can/will verify the server's certificate is
+     * not the server's concern. The client should never reveal this; varying
+     * it effectively tells a MITM attacker whether or not the client will be
+     * able to *detect* a MITM attack based on 2-sided TLS.
+     *
+     * We set this bit to OFF, not to ON, in order to make our ouput
+     * indistinguishable from older clients which vary this bit, so that those
+     * older clients become more difficult for MITM attackers to reliably target
+     * once this change is deployed.
+     */
+    ulong sanitized_client_flag= mysql->client_flag & ~CLIENT_SSL_VERIFY_SERVER_CERT;
+
     /* 4.1 server and 4.1 client has a 32 byte option flag */
     if (!(mysql->server_capabilities & CLIENT_MYSQL))
       mysql->client_flag&= ~CLIENT_MYSQL;
-    int4store(buff,mysql->client_flag);
+
+    int4store(buff, sanitized_client_flag);
     int4store(buff+4, net->max_packet_size);
     buff[8]= (char) mysql->charset->nr;
     memset(buff + 9, 0, 32-9);


dlenski:
After I created [CONC-656](https://jira.mariadb.org/browse/CONC-656), @9E0R9 pushed https://github.com/mariadb-corporation/mariadb-connector-c/commit/15e6b54a9babbe4bcc532ec4c33254cb5ef888d5, which also addresses the same issue, but without explaining _why_ it's important to address this issue.

In turn, that commit https://github.com/mariadb-corporation/mariadb-connector-c/commit/15e6b54a9babbe4bcc532ec4c33254cb5ef888d5 introduced a very serious new security bug, which I explained [in a comment on it](https://github.com/mariadb-corporation/mariadb-connector-c/commit/15e6b54a9babbe4bcc532ec4c33254cb5ef888d5#r121441048).

The next day, that commit was then quietly replaced with https://github.com/mariadb-corporation/mariadb-connector-c/commit/45feebb99dda104a0f12abfba53ba90eabc87ff7, which fixes the security bug in precisely the way I described.

- Nowhere is the security importance of this fix mentioned
- Nowhere am I credited for discovering and fixing the security vulnerability
- Nowhere am I credited for pointing out the severe new security bug introduced by @9EOR9's first attempt at an alternative fix

9EOR9:
SSL/TLS support was added in MySQL 4.0 on top of the pre-4.1 protocol, where the TLS handshake had to be integrated in the client/server handshake.
This limitation was necessary not only for backwards compatibility, but also due to the fact that the MySQL client/server protocol starts with a "server hello" packet, while TLS handshake starts with a "client hello" packet. This is of course not state of the art, and we are currently discussing how to solve this (e.g. server listens on an additional TLS only port).

A MitM could spy the capability flags and detect which features are supported or not, e.g. if the client will validate the server certificate.
If this information is not available in unencrypted part of the handshake, the MitM can detect this itself:
- by checking TLS client alert codes (in case cert verification will be handled during handshake)
```
4111	4191.108722	127.0.0.1	127.0.0.1	TLSv1.3	204	Server Hello
4113	4191.110578	127.0.0.1	127.0.0.1	TLSv1.3	72	Application Data
4115	4191.110598	127.0.0.1	127.0.0.1	TLSv1.3	109	Application Data
4117	4191.110839	127.0.0.1	127.0.0.1	TLSv1.3	1768	Application Data
4119	4191.123055	127.0.0.1	127.0.0.1	TLSv1.3	51	Alert (Level: Fatal, Description: Unknown CA)
```

- or by checking if the 2nd packet after handshake completion is a FIN/ACK packet from client:

```
129	90.903985	::1	::1	TLSv1.2	70	Change Cipher Spec
131	90.904057	::1	::1	TLSv1.2	109	Encrypted Handshake Message
--- end of TLS handshake ---
132	90.904072	::1	::1	TCP	64	61414 → 3306 [ACK] Seq=335 Ack=2702 Win=324608 Len=0
133	90.992828	::1	::1	TCP	64	61414 → 3306 [FIN, ACK] Seq=335 Ack=2702 Win=324608 Len=0
```

If MitM uses a signed certificate with the correct hostname, it can also be determined if the client verificates the fingerprint of server certificate.

So exposing the CLIENT_SSL_VERIFY_SERVER_CERT flag is **not** a security risk, since this information can be easily obtained as shown above.

In our team meeting in June Serg and me decided to remove the CLIENT_SSL_VERIFY_SERVER_CERT both from server and client, since it is a client only option, but not a capability. This "bug" was added in 2006 (introduced by server commit 89d759b93e3975e5d5e1c5cf9b901c01b9e80ff7) and fixed in latest server and Connector/C releases.
The client side fix was corrected and forced pushed after getting travis/buildbot bug results. I can't tell you anymore if this happened in my working branch - if it was in 3.3 branch by mistake, then I appologize. This should of course never happen.

vuvova:
@dlenski, I'm closing this PR. I like your fix more, this is how the bug should've been fixed, but @9EOR9's fix, while redundantly more complex, is technically not wrong. And he is the maintainer of the project, so I don't see a good reason for redoing an already pushed fix.

PR #229, opened 2023-08-16
CONC-659: Fix warnings

markus456:
The code that causes the warnings is what actually caused the problems. This commit fixes the warnings and turns on -Werror which should prevent this from happening in the future.


markus456:
Closing this as the fixes were already integrated into the 3.3 branch.

PR #231, opened 2023-09-08
MDEV-19511 Remove WORDS_BIGENDIAN - HAVE_BIGENDIAN replaced it

grooverdan:
Also note https://cmake.org/cmake/help/latest/module/TestBigEndian.html

Discussion source, https://lists.launchpad.net/maria-developers/msg11819.html

It was solved a while ago with HAVE_BIGENDIAN added to test, WORDS_BIGENDIAN wasn't cleaned up.


grooverdan:
replace by 3.1 PR.

PR #232, opened 2023-09-12
3.3 Refactor zlib functions to use modern prototype declaration and avoid warnings.

joaquinbejar:
Refactor zlib functions to use modern prototype declaration and avoid warnings.

Updated the declaration of the `zlib` functions to use modern C++ syntax for specifying parameter types. This resolves a compiler warning about using an outdated function declaration style.

New Signature:
`uLong ZEXPORT adler32_z(uLong adler, const Bytef *buf, z_size_t len)`

This change improves code compatibility with modern C/C++ standards.

### DIFF
--- a/external/zlib/adler32.c
+++ b/external/zlib/adler32.c
@@ -13,10 +13,10 @@ local uLong adler32_combine_ OF((uLong adler1, uLong adler2, z_off64_t len2));
 #define NMAX 5552
 /* NMAX is the largest n such that 255n(n+1)/2 + (n+1)(BASE-1) <= 2^32-1 */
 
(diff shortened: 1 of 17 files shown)


9EOR9:
Thank you for your pull request! 

Unfortunately we can't merge your pull request since we frequently update our sources from zlib project (and in future probably will add zlib as a git submodule).  Without extra patching (which causes a lot of extra work) your changes would be lost after next zlib update.

So it would be great if you could file a PR in the [zlib github project](https://github.com/madler/zlib)!

joaquinbejar:
Hi @9EOR9,

thx for the reply I'll continue on zlib repo

PR #234, opened 2023-09-13
Compile with MinGW

Iswenzz:
Compiles with MinGW 13.1 when disabling WARNING_AS_ERROR.


9EOR9:
Hi,

could you please check if fix for [CONC-645](https://jira.mariadb.org/browse/CONC-645) solves this issue? Unfortunately this task was incorrectly filed under CONJ (Java Connector) first, and was therefore not on my radar.

PR #236, opened 2023-09-23
Update ma_context.c

tildeslash:
Old K&R function declaration is no longer supported when compiling on macOS 13.5.2 with clang 15.x.


9EOR9:
Thanks for your PR.

Could you please change the branch to 3.1 (Server 10.4 and 10.5 still use old 3.1 branch) - so we can merge afterwards into 3.3

PR #238, opened 2023-10-21
ARM 32-bit platforms are getting a bad handshake error

wagnerch:
This issue does not occur on 3.3.6, and appears to be caused by commit https://github.com/mariadb-corporation/mariadb-connector-c/commit/9f37c27bc8921ddc7e65ba8fc75cb4993380228a.

The solution that seems to work on arm & aarch64 is casting length to ulonglong before calling int8store.  Runtime tested on arm & aarch64 platforms.

### DIFF
--- a/libmariadb/mariadb_lib.c
+++ b/libmariadb/mariadb_lib.c
@@ -123,7 +123,7 @@ extern int mthd_stmt_fetch_to_bind(MYSQL_STMT *stmt, unsigned char *row);
 extern int mthd_stmt_read_all_rows(MYSQL_STMT *stmt);
 extern void mthd_stmt_flush_unbuffered(MYSQL_STMT *stmt);
 extern my_bool _mariadb_read_options(MYSQL *mysql, const char *dir, const char *config_file, const char *group, unsigned int recursion);
-extern unsigned char *mysql_net_store_length(unsigned char *packet, ulonglong length);
+extern unsigned char *mysql_net_store_length(unsigned char *packet, size_t length);
 
 extern void
 my_context_install_suspend_resume_hook(struct mysql_async_context *b,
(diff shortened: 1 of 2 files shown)


wagnerch:
Assuming mysql_net_store_length is private then another option is to drop this patch and update the types in the authentication plugin.  This also resolves the issue on 32-bit, which I guess is caused by passing in a 32-bit integer to a function spec that is expecting a 64-bit integer.

```
diff --git a/libmariadb/ma_net.c b/libmariadb/ma_net.c
index 35301090a522..fbf43963ecef 100644
--- a/libmariadb/ma_net.c
+++ b/libmariadb/ma_net.c
@@ -292,8 +292,6 @@ static int ma_net_write_buff(NET *net,const char *packet, size_t len)
   return 0;
 }

-unsigned char *mysql_net_store_length(unsigned char *packet, size_t length);
-
 /*  Read and write using timeouts */

 int ma_net_real_write(NET *net, const char *packet, size_t len)
diff --git a/plugins/auth/my_auth.c b/plugins/auth/my_auth.c
index f9c937689e73..8cfce3defabc 100644
--- a/plugins/auth/my_auth.c
+++ b/plugins/auth/my_auth.c
@@ -12,7 +12,7 @@ static int dummy_fallback_auth_client(MYSQL_PLUGIN_VIO *vio, MYSQL *mysql __attr
 extern void read_user_name(char *name);
 extern char *ma_send_connect_attr(MYSQL *mysql, unsigned char *buffer);
 extern int ma_read_ok_packet(MYSQL *mysql, uchar *pos, ulong length);
-extern unsigned char *mysql_net_store_length(unsigned char *packet, size_t length);
+extern unsigned char *mysql_net_store_length(unsigned char *packet, ulonglong length);

 typedef struct {
   int (*read_packet)(struct st_plugin_vio *vio, uchar **buf);
```

9EOR9:
Thanks for your PR. I think this was fixed already with 8320f0d54d50635e82b723c151ac92c9528d0e79 (3.1 branch).

wagnerch:
Thanks @9EOR9, yes that would fix it.

PR #239, opened 2023-11-03
Add support to ZSTD as static library

uilianries:
The ZSTD library uses `zstd_static` as name when is static:

https://github.com/facebook/zstd/blob/dev/build/cmake/lib/CMakeLists.txt#L161

Without this change, is not possible to find zstd on Windows, for instance, because the current `FindZstd.cmake` expects `zstd` as library name only.

### DIFF
--- a/cmake/FindZStd.cmake
+++ b/cmake/FindZStd.cmake
@@ -10,7 +10,7 @@ find_path(ZSTD_INCLUDE_DIRS
   HINTS ${ZSTD_ROOT_DIR}/include)
 
 find_library(ZSTD_LIBRARIES
-  NAMES zstd
+  NAMES zstd zstd_static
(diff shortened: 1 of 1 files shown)


9EOR9:
Thanks for your contribution. Fix will be available in version 3.3.9

cmake/FindZStd.cmake
```diff
@@ -10,7 +10,7 @@ find_path(ZSTD_INCLUDE_DIRS
   HINTS ${ZSTD_ROOT_DIR}/include)
 
 find_library(ZSTD_LIBRARIES
-  NAMES zstd
+  NAMES zstd zstd_static
```
dg0yt, line 13: Whenever using multiple names, use `NAMES_PER_DIR`.

PR #242, opened 2024-03-04
[MDEV-30366] Bulk unitary result flag client implementation part.

rusher:
With MDEV-30366, server now permit to send a result-set containing generated id and Affected rows for each bulk operation. This feature can be enabled with option MARIADB_OPT_BULK_UNIT_RESULTS when server supports it.

### DIFF
--- a/libmariadb/mariadb_lib.c
+++ b/libmariadb/mariadb_lib.c
@@ -693,6 +693,7 @@ struct st_default_options mariadb_defaults[] =
   {{MYSQL_OPT_SSL_ENFORCE}, MARIADB_OPTION_BOOL, "ssl-enforce"},
   {{MARIADB_OPT_RESTRICTED_AUTH}, MARIADB_OPTION_STR, "restricted-auth"},
   {{.option_func=parse_connection_string}, MARIADB_OPTION_FUNC, "connection"},
+  {{MARIADB_OPT_BULK_UNIT_RESULTS}, MARIADB_OPTION_BOOL, "bulk-unit-results"},
   /* Aliases */
   {{MARIADB_OPT_SCHEMA}, MARIADB_OPTION_STR, "db"},
   {{MARIADB_OPT_UNIXSOCKET}, MARIADB_OPTION_STR, "unix_socket"},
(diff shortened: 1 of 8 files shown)


libmariadb/mariadb_lib.c
```diff
@@ -3830,6 +3832,9 @@ mysql_optionsv(MYSQL *mysql,enum mysql_option option, ...)
       }
     }
     break;
+  case MARIADB_OPT_BULK_UNIT_RESULTS:
+    mysql->options.bulk_unit_results= *(my_bool *)arg1;
```
9EOR9: This will break ABI, since the size of MYSQL will change. Intstead option should be stored in mysql->options.extensions via OPT_SET_EXTENDED_VALUE (and bulk_unit_results should be moved from st_mysql_options to st_mysql_options_extension in ma_common.h)

unittest/libmariadb/bulk1.c
```diff
@@ -21,6 +21,13 @@
 
 static my_bool bulk_enabled= 0;
 
+#define SERVER_SUPPORT_BULK_UNIT_RESULTS(stmt)\
```
9EOR9: better to pass mysql instead of stmt to the macro, so it could be used without creating a statement, e.g. to skip a test .

unittest/libmariadb/bulk1.c
```diff
@@ -1067,6 +1077,345 @@ static int test_mdev16593(MYSQL *mysql)
...
+  MYSQL_BIND bind[1];
+  MYSQL_RES *res;
+  MYSQL_ROW row;
+  MYSQL_BIND bind_out[2];
+  int id, affected_rows = 0;
+  int expectedId = 1;
+  unsigned int intval;
+
+  SKIP_MAXSCALE;
+
```
9EOR9, line 1104:
```
if (!SERVER_SUPPORT_BULK_UNIT_RESULTS(my))
{
  diag("Server doesn't support bulk unit results");
  return SKIP;
}
```

9EOR9 (commented):
review finished (see comments above)

PR #243, opened 2024-03-27
CONC-693 Fix SSL_read/write return value checking in ma_tls_async_check_result

joshuahunt:
SSL_{read,write}'s return values == 0 signify the operation was unsuccessful, but here it's being treated as success. Other calls of these functions already properly checks the return value.

Please note I found this through code inspection while trying to debug a different issue. I do not have a case where this can be hit, but in principle this should be fixed. Link to openssl man page showing return values <= 0 signify an error https://www.openssl.org/docs/manmaster/man3/SSL_read.html

### DIFF
--- a/libmariadb/mariadb_async.c
+++ b/libmariadb/mariadb_async.c
@@ -140,7 +140,7 @@ my_ssl_async_check_result(int res, struct mysql_async_context *b, MARIADB_SSL *c
 {
   int ssl_err;
   b->events_to_wait_for= 0;
-  if (res >= 0)
+  if (res > 0)
     return 1;
   ssl_err= SSL_get_error(ssl, res);
   if (ssl_err == SSL_ERROR_WANT_READ)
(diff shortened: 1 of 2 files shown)


joshuahunt:
It looks like this commit fixed the problem for `ma_tls_read()` but missed it for `ma_tls_async_check_result()` @9E0R9 
```
commit f1b08b836931daf5b20405b567a6dee2d6a89400
Author: Georg Richter <georg@mariadb.com>
Date:   Fri Jul 8 07:46:00 2022 +0200

    Partial fix for MDEV-27405:
    
    The return value of SSL_read indicates an error if it is <= 0, not
    if it is < 0.
```

9EOR9:
Thanks for your pull request! Could you please provide the PR against 3.1  branch (still included in Server 10.5), I will then merge it from 3.1 into 3.4 (3.3 will be disontinued).

joshuahunt:
> Thanks for your pull request! Could you please provide the PR against 3.1 branch (still included in Server 10.5), I will then merge it from 3.1 into 3.4 (3.3 will be disontinued).

OK I rebased it on top of 3.1. Please let me know if there's anything else you need me to do.

PR #245, opened 2024-04-25
[MDEV-34009] Client-side implementation of instant failover mechanism (and TLS fixes)

dlenski:
This PR is a replacement for https://github.com/mariadb-corporation/mariadb-connector-c/pull/226, which proposed a very similar feature under the name of "server redirect.

Upstream developers chose a different approach (see https://github.com/mariadb/server/pull/1472 and https://github.com/mariadb-corporation/mariadb-connector-c/pull/137) for that feature. See [MDEV-15935](https://jira.mariadb.org/browse/MDEV-15935) ("connection redirection") for further discussion.

I maintain that a simple and obligatory server-directed failover mechanism will be a very useful feature for many users, so I'm reintroducing it here under the alternative name of "instant failover."

---

This is the client counterpart of https://github.com/mariadb/server/pull/3224, which provides an initial server-initiated mechanism for instantly redirecting new client connections to a failover server.

It modifies the Connector/C library, so that if the server responds to our initial connection with an error packet having
error number `ER_INSTANT_FAILOVER`), and an error string formatted as `|Human-readable message|host[:port]`, the client will redirect its connection accordingly.

This redirection is performed in `mthd_my_real_connect`, which appears to be the most appropriate location given that it already contains code for retrying/repeating a new server connection.


vuvova:
We already have a failover mechanism, we aren't going to do another one

PR #248, opened 2024-06-15
mariadb_config: cmake: Don't prefix `/absolute/file.a` with `-l`

nh2:
Because `-l` does not take absolute paths.

Context:
https://github.com/mariadb-corporation/mariadb-connector-c/commit/8aa86f73adcab01147416e4c5f4876532da4bd4b#commitcomment-143167684


robinnewhouse:
Just reviewing some open PRs. This looks like a good change to me.

PR #251, opened 2024-08-12
MDEV-34730: disable TLS over UNIX socket

erkia:
Proposing a fix to https://jira.mariadb.org/browse/MDEV-34730

Using TLS over UNIX socket adds no extra security while causing problems when using real valid certificates, which are not covered by tls_self_signed_error checks and therefore fail the verification over UNIX socket.


erkia:
Can't blindly disable TLS for UNIX socket.

PR #256, opened 2024-09-25
Fix ushort type check

dg0yt:
Fixes wrong check result. Fixes Android build.


dg0yt:
Fixes
~~~
/vcpkg/vcpkg/buildtrees/libmariadb/src/5316b9c3c2-a2a62641d5.clean/plugins/auth/my_auth.c:186:21: error: use of undeclared identifier 'ushort'; did you mean 'qsort'?
  186 |     int2store(end, (ushort) mysql->charset->nr);
      |                     ^~~~~~
      |                     qsort

PR #259, opened 2024-10-26
CONC-738: Fix zstd compression level bytes

markus456:
The MySQL protocol documentation as well as the comments in code state that the compression level is stored in one byte. The code, however, used four bytes.


markus456:
Fixed the commit to conform to the coding style.

PR #261, opened 2024-11-07
Add configurable zstd compression level

markus456:
The compression level has been hard-coded to the default value of 3. Making it configurable makes the connector-c more compatible with MySQL which has configurable compression levels.


markus456:
This depends on pull request #259 being merged before this one.

PR #263, opened 2024-12-11
CONC-749 plugins.multiauth fails on mac

DaveGosselin-MariaDB:
If the dylib isn't found on linux, then dlerror() returns 'no such file' but on Mac it returns a lot more information. Our tests expect the string 'no such file', so detect that error on Mac and replace the text.

### DIFF
--- a/libmariadb/ma_client_plugin.c.in
+++ b/libmariadb/ma_client_plugin.c.in
@@ -352,6 +352,22 @@ mysql_client_register_plugin(MYSQL *mysql,
 }
 
 
+#ifdef __APPLE__
+/*
+  If the dylib isn't found on linux, then dlerror() returns 'no such file'
+  but on Mac it returns a lot more information.  Our tests expect the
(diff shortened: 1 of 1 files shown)


vuvova:
"a lot more information" is usually good. If it's a useful information. Better to adjust the test to be less picky

DaveGosselin-MariaDB:
> "a lot more information" is usually good. If it's a useful information. Better to adjust the test to be less picky

It's not really useful, just path information, which isn't available on Linux.

9EOR9:
I would keep this information. I don't know if multiauth is skipped or limited on windows, but we also extend the Windows system error message with [path information](https://github.com/mariadb-corporation/mariadb-connector-c/blob/3.4/libmariadb/ma_client_plugin.c.in#L408)

DaveGosselin-MariaDB:
I'm happy to keep it, but I'm wondering how to make the test resilient to the differences in output across operating systems, do you have advice?

PR #265, opened 2024-12-31
Avoid redfine of IO_SIZE if made configurable

ericherman:
**This PR Closed and replaced by: https://github.com/mariadb-corporation/mariadb-connector-c/pull/267**

If the server is to make IO_SIZE configurable, we need to avoid either redefining it, or defining it to be different.

See: https://github.com/MariaDB/server/pull/3726 and https://jira.mariadb.org/browse/MDEV-35740

### DIFF
--- a/include/ma_global.h
+++ b/include/ma_global.h
@@ -433,12 +433,6 @@ typedef SOCKET_SIZE_TYPE size_socket;
 /* #define FN_UPPER_CASE TRUE */
 
 /*
-  Io buffer size; Must be a power of 2 and a multiple of 512. May be
-  smaller what the disk page size. This influences the speed of the
-  isam btree library. eg to big to slow.
-*/
-#define IO_SIZE			4096
-/*
   How much overhead does malloc have. The code often allocates
   something like 1024-MALLOC_OVERHEAD bytes
 */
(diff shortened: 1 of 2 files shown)


include/ma_global.h
```diff
@@ -433,11 +433,17 @@ typedef SOCKET_SIZE_TYPE size_socket;
...
+  This influences the speed of the isam btree library. E.g.: too big too slow.
+  4096 is a common block size on SSDs.
 */
-#define IO_SIZE			4096
+#ifndef IO_SIZE
+#define IO_SIZE			4096U
+#endif
+#if ((IO_SIZE <= 0) || ((IO_SIZE % 512) != 0) || ((IO_SIZE & (IO_SIZE-1)) != 0))
+#error "IO_SIZE must be a positive multiple of 512 and power of 2"
+#endif
```
vuvova: I don't like that, there's no reason — that I can see — why server's IO_SIZE should affect the client. I'd suggest to rename IO_SIZE in the libmariadb. See how it's used, perhaps a `NET_BUFF_ALIGN` could be a good name

vuvova (changes requested):

include/ma_global.h
ericherman: I would be perfectly happy to rename client constant instead as I also feel like these should be able to be distinct; @9E0R9 what do you think?

ericherman: I've updated and --force pushed

vuvova (approved):
look good to me

PR #267, opened 2025-01-05
Create #define for network buffer alignment

ericherman:
Was: https://github.com/mariadb-corporation/mariadb-connector-c/pull/265

The server's definition of IO_SIZE is re-used here in the client for network buffer alignment, however IO_SIZE is used in the server for many different things and the client's buffer alignment is not related to many of those uses.

If the server is to make IO_SIZE configurable, we need to avoid either redefining it, or defining it to be different. By creating a specific define for this, we avoid redfine and clarify the code.

See: https://github.com/MariaDB/server/pull/3726

### DIFF
--- a/include/ma_global.h
+++ b/include/ma_global.h
@@ -413,12 +413,6 @@ typedef SOCKET_SIZE_TYPE size_socket;
 /* #define FN_UPPER_CASE TRUE */
 
 /*
-  Io buffer size; Must be a power of 2 and a multiple of 512. May be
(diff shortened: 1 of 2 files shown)


ericherman:
@vuvova , I imagine your comment ( https://github.com/mariadb-corporation/mariadb-connector-c/pull/265#pullrequestreview-2530864995 ) still stands, yes?

vuvova:
Yes, it does

grooverdan (approved):

PR #271, opened 2025-02-11
CONC-755: unit.conc_ps_bugs test (test_conc623)

grooverdan:
Initialize bind structure to prevent MSAN error.


grooverdan:
fixed via 28ae227e801fd9ad5f5589658dad872060854a9b

PR #273, opened 2025-04-01
CONC-764: Build error in ma_context.c on android

knielsen:
X18 is a platform-reserved register on Android, not a callee-save register. So it will not be touched by the spawned/resumed co-routine and must not be included in the GCC asm clobber list on this platform.

After some research, it looks like X18 is a scratch register on Linux, but platform reserved on most other platforms, including Android, MacOS, FreeBSD, Windows. So the patch puts X18 in the clobber list on Linux only (and not on Android, which also defines the `__linux__` preprocessor macro).

References:

Build failure: https://github.com/microsoft/vcpkg/pull/44015
Android arm64-v8a ABI: https://developer.android.com/ndk/guides/abis#arm64-v8a


martinp2812:
The same problem is present on MacOS on ARM64:

```
   ext/mariadb-connector-c-3.4.4-src/libmariadb/ma_context.c:676:6: error: inline asm clobber list contains reserved registers: X18 [-Werror,-Winline-asm]
     676 |      "mov x10, sp\n\t"
         |      ^
   ext/mariadb-connector-c-3.4.4-src/libmariadb/ma_context.c:676:6: note: Reserved registers on the clobber list may not be preserved across the asm statement, and clobbering them may lead to undefined behaviour.
   ext/mariadb-connector-c-3.4.4-src/libmariadb/ma_context.c:746:6: error: inline asm clobber list contains reserved registers: X18 [-Werror,-Winline-asm]
     746 |      "ldp x13, x11, [%[save], #0]\n\t"
         |      ^
   ext/mariadb-connector-c-3.4.4-src/libmariadb/ma_context.c:746:6: note: Reserved registers on the clobber list may not be preserved across the asm statement, and clobbering them may lead to undefined behaviour.
   ext/mariadb-connector-c-3.4.4-src/libmariadb/ma_context.c:846:6: error: inline asm clobber list contains reserved registers: X18 [-Werror,-Winline-asm]
     846 |      "ldp x13, x11, [%[save], #0]\n\t"
         |      ^
   ext/mariadb-connector-c-3.4.4-src/libmariadb/ma_context.c:846:6: note: Reserved registers on the clobber list may not be preserved across the asm statement, and clobbering them may lead to undefined behaviour.
   3 errors generated.
```

knielsen:
Interesting. And it seems X18 is also a platform reserved register on arm64 Windows.
Maybe the \#ifdef should then be changed to only include X18 in the clobber list on Linux? Though I'll need to check the *BSD's too.

knielsen:
This is quite annoying, actually :-/

The libmariadb code in ma_context.cc does not use the X18 register in any way, it will work fine on any of the platforms.

On platforms where X18 is reserved, the ma_context.cc code will also preserve it. On other platforms, the X18 register need not be preserved by the ma_context.cc code.

The only case to handle is to tell GCC itself to not generate code that uses X18 in the function, which is only relevant on platforms where it is not reserved.

In summary, the code is correct in all cases, it's just the compiler warning that is bogus. And -Werror is wrongly causing a compile failure here.

tobil4sk:
Omitting this register from the list fixes our mac arm64 build of mariadb-connector-c

PR #274, opened 2025-04-03
CONC-766: Disable clang -Wcast-function-type-strict for makecontext

grooverdan:
makecontext has a defined prototype in ucontext.h that differs from its expected usage. Disable the clang warning for this function call.

Modelled off the af4498b776b1703a00cbb4f86ead0d0e4f5332bb commit.

### DIFF
--- a/libmariadb/ma_context.c
+++ b/libmariadb/ma_context.c
@@ -103,8 +103,21 @@ my_context_spawn(struct my_context *c, void (*f)(void *), void *d)
   c->user_data= d;
   c->active= 1;
   u.p= c;
+  /*
+    makecontext function expects function pointer to receive multiple
+    ints as an arguments, however is declared in ucontext.h header with
+    a void (empty) argument list. Ignore clang cast-function-type-strict
+    warning for this function call.
+  */
+# ifdef __clang__
+#  pragma clang diagnostic push
+#  pragma clang diagnostic ignored "-Wcast-function-type-strict"
+# endif
   makecontext(&c->spawned_context, (uc_func_t)my_context_spawn_internal, 2,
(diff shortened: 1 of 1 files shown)


knielsen (approved):
The patch is ok.

An alternative might be to build with cmake arg `-DWITH_BOOST_CONTEXT=ON`. Then it will not try to use ucontext, even in address sanitizer build.

You might want to also remove this comment earlier in the file. Someone "Tildeslash" changed the code to remove the original fix for this function type mismatch without even bothering to fix (or even read?) the comment :-( :
```c
/*
  We use old-style function definition here, as this is passed to
  makecontext(). And the type of the makecontext() argument does not match
  the actual type (as the actual type can differ from call to call).
*/
```

grooverdan:
> The patch is ok.

Thank you.

> An alternative might be to build with cmake arg `-DWITH_BOOST_CONTEXT=ON`. Then it will not try to use ucontext, even in address sanitizer build.

Rather stick to testing as close the code that is released as possible.

> You might want to also remove this comment earlier in the file. Someone "Tildeslash" changed the code to remove the original fix for this function type mismatch without even bothering to fix (or even read?) the comment :-( :
> 
> ```c
> /*
>   We use old-style function definition here, as this is passed to
>   makecontext(). And the type of the makecontext() argument does not match
>   the actual type (as the actual type can differ from call to call).
> */
> ```

So belatedly maybe:

```c
/*
   Note, the prototype of makecontext doesn't match the calling convention as
   the types of arguments can differ from call to call. Because of this, the warning 
  `Wcast-function-type-strict` is disabled below around setting up the callback to
  this function.
*/
```

PR #276, opened 2025-05-13
Fix my_context_[spawn|continue|yield] for aarch64 BTI

stewartsmith:
Fixes `SIGILL` in MariaDB Server `main.non_blocking_api` test case on aarch64 processors supporting Branch Target Identification (BTI).

Branch Target Identification is a new aarch64 feature that prevents jumping to arbitrary code locations that are not properly marked as branch targets.

The `-mbranch-protection=standard` flag to GCC will enable code generation using BTI. This is backwards compatible as the added instructions are no-ops on older processors.

One example of where both `-mbranch-protection=standard` is enabled and hardware supports it is with Amazon Linux 2023 and Graviton 4.

The symptom is the following mtr `main.non_blocking_api` failure:
```
  mysqltest got signal 4
  read_command_buf (0xaaaae6ae27a8): connect
  (con_nonblock,localhost,root,,test)

  conn->name (0xaaaae6b0c0b8):

  Attempting backtrace...
  stack_bottom = 0x0 thread_stack 0x3c000
  /usr/bin/mariadb-test(my_print_stacktrace+0x48)[0xaaaab6934088]
  bits/stdio2.h:79(signal_handler)[0xaaaab68f41a0]
  addr2line: 'linux-vdso.so.1': No such file
  linux-vdso.so.1(__kernel_rt_sigreturn+0x0)[0xffffa22bc830]
  /usr/bin/mariadb-test(my_context_spawn+0x74)[0xaaaab6914d84]
```
With GDB giving the information that SIGILL was on the perfectly legal 'mov' instruction:
```
  Program received signal SIGILL, Illegal instruction.
  0x0000aaaaaab44d84 in my_context_spawn (c=0xaaaaab07ada8, f=0xfffff7fed240, d=0x0) at
  mariadb1011-10.11.11-1.amzn2023.0.1.aarch64/libmariadb/libmariadb/ma_context.c:674
  674       __asm__ __volatile__
  (gdb) x/i$pc
  => 0xaaaaaab44d84 <my_context_spawn(my_context*, void (*)(void*),
  void*)+116>: mov      w0, #0x1                        // #1
```
The call sequence to get to this point is as follows:
```
  Breakpoint 1, my_context_spawn (c=0xaaaaab07ada8, f=0xaaaaaab38024
  <mysql_real_connect_start_internal(void*)>, d=0xffffffffc648) at
  libmariadb/ma_context.c:661
  661       register void *stack asm("x13") = c->stack_top;
  (gdb) c
  Continuing.

  Breakpoint 2, my_context_yield (c=0xaaaaab07ada8) at
  libmariadb/ma_context.c:843
  843       register const uint64_t *save asm("x19") = &c->save[0];

  Continuing from here leads to the SIGILL above.
```
The branch that ends up causing the SIGILL is the `br x11` instruction in `my_context_yield()`. This is due to the branch being to a location that is not marked as a branch target.

The fix is to ensure that these locations in the `my_context` aarch64 assembler are marked with `bti j` instructions so that the locations that will be `br`ed to are valid to be `br`ed to.

Since older compilers don't know about the instruction, we need an `ifdef` to define a string that has the raw opcode in it.

Fixes: d2285fb8304c82fd87de483845d1c12b06927479 (introduced in v3.3.12)

---
**NOTE**: This will need to make its way into MariaDB Server as well in order for the fix to appear there.

### DIFF
--- a/libmariadb/ma_context.c
+++ b/libmariadb/ma_context.c
@@ -637,6 +637,13 @@ my_context_destroy(struct my_context *c)
 
 
 #ifdef MY_CONTEXT_USE_AARCH64_GCC_ASM
+
+#if (defined(__clang__) && (__clang_major__ >= 9)) ||                   \
+        (defined(__GNUC__) && (__GNUC__ > 9 || (__GNUC__ == 9 && __GNUC_MINOR__ >= 1)))
+#define BTI_J_STR "bti j"
+#else
+#define BTI_J_STR ".inst 0xd503241f"
+#endif
 /*
   GCC-aarch64 (arm64) implementation of my_context.
 
@@ -724,12 +731,14 @@ my_context_spawn(struct my_context *c, void (*f)(void *), void *d)
        will do this for us if needed.
      */
      "1:\n\t"
+     BTI_J_STR "\n\t"
      "ldr x10, [%[save], #88]\n\t"
      "mov sp, x10\n\t"
      "mov %w[ret], #0\n\t"
      "b 3f\n"
      /* Come here when operation was suspended. */
      "2:\n\t"
+     BTI_J_STR "\n\t"
      "mov %w[ret], #1\n"
      "3:\n"
      : [ret] "=r" (ret),
@@ -824,6 +833,7 @@ my_context_continue(struct my_context *c)
        my_context_spawn(), so we preserve the value correctly at this point.
      */
      "1:\n\t"
+     BTI_J_STR "\n\t"
      /* x19 (aka %[save]) is preserved from my_context_spawn() in this case. */
      "ldr x20, [%[save], #8]\n\t"
      "ldp x21, x22, [%[save], #16]\n\t"
@@ -840,6 +850,7 @@ my_context_continue(struct my_context *c)
      "b 3f\n"
      /* Come here when operation is suspended. */
      "2:\n\t"
+     BTI_J_STR "\n\t"
      "mov %w[ret], #1\n"
      "3:\n"
      : [ret] "=r" (ret)
@@ -923,6 +934,7 @@ my_context_yield(struct my_context *c)
      "br x11\n"
 
      "1:\n"
+     BTI_J_STR "\n\t"
      :
      : [save] "r" (save)
      : "x0", "x1", "x2", "x3", "x4", "x5", "x6", "x7",


stewartsmith:
I've also filed the following JIRA issues:
- https://jira.mariadb.org/browse/CONC-774 for the client library
- https://jira.mariadb.org/browse/MDEV-36807 for the server

grooverdan:
Thanks @stewartsmith, long time no see :smile:

knielsen:
Thanks, Stewart. The patch looks good.

Did you consider making the added BTI instructions conditional on whether the -mbranch-protection=standard is in the CFLAGS? After all, adding this makes little sense if the rest of libmariadb is compiled without the added BTI's. But it might be simpler to just do it unconditionally as in this patch, instead of trying to get the checks right and getting a long tail of corner cases not caught.oO

I'm sad to see needless complexity like this BTI creeping in :-( . But the non-blocking API can do little to help with that problem.

A work-around for this issue in versions without this patch is to compile with `-DWITH_BOOST_CONTEXT=ON` (assuming this is fixed in boost::context, I have not checked).

9EOR9:
Hi Stewart, long time not seen! Thanks for your contribution!

stewartsmith:
Thanks for the quick review and merge! Good to see everyone again :)

The part where there's a SIGILL on a perfectly valid instruction was the head-scratching/annoying part for debugging. I'm hoping there's some nice debug register somewhere that helps make that easier.

I ended up looking into Boost::context (at least the version we ship in AL2023), and I couldn't get it to do something wrong in some simple tests, but I also ended up scratching my head a bit too much trying to understand the source code for it and just tried to prove it correct with code instead.

libmariadb/ma_context.c
```diff
@@ -637,6 +637,13 @@ my_context_destroy(struct my_context *c)
 
 
 #ifdef MY_CONTEXT_USE_AARCH64_GCC_ASM
+
+#if (defined(__clang__) && (__clang_major__ >= 9)) ||                   \
+        (defined(__GNUC__) && (__GNUC__ > 9 || (__GNUC__ == 9 && __GNUC_MINOR__ >= 1)))
+#define BTI_J_STR "bti j"
+#else
+#define BTI_J_STR ".inst 0xd503241f"
+#endif
```
dr-m, line 646: GCC will jut pass through the `__asm__` code to the underlying assembler. To my knowledge, there is no preprocessor directive that would identify the assembler version. On a Ubuntu 20.04 builder, this is causing failures like this:
```
/tmp/cc7iCqmP.s:56: Error: selected processor does not support `bti j'
```
I think that this needs to be revised so that we only check for Clang, which features a built-in assembler. Anywhere else, we’d use the numeric encoding, to be compatible with an older assembler.

dr-m: I tested this on https://godbolt.org:
```c++
// Type your code here, or load an example.
int square(int num) {
    __asm__ __volatile__("bti j");
    return num * num;
}
```
The oldest armv8 clang version I found was clang 9.0.0. It would not accept this unless I specify `-march=armv8.5-A` or something like that. Only starting with clang 12.0 the code would compile with the default settings. So, I think that this needs to be revised to the following:
```c
#if defined __clang_major__ && __clang_major__ >= 12
# define BTI_J_STR "bti j"
#else
# define BTI_J_STR ".inst 0xd503241f"
#endif
```
Curiously, in my test on https://godbolt.org the encoding is ` 0xd503249f` instead of ` 0xd503241f`. What is the difference? In https://godbolt.org one can see the encoding by enabling "Compile to binary code" in the output settings (the gear symbol).

Actually, the assembler on https://godbolt.org will fail to recognize this instruction as well, even if I try to play with `-march`: https://godbolt.org/z/948PdGeT7

dr-m: According to https://www.spinics.net/lists/arm-kernel/msg1041915.html the two most significant bits of the least significant octet of the instruction word are parameters. The inline assembler in my clang experiment above actually inserted  `bti j`, but the numeric encoding in @stewartsmith’s patch is for a `bti` instruction without any `j`, `c`, or `jc` argument. If we actually want `bti j`, it needs to be `0xd503249f`.

PR #287, opened 2025-06-19
mariadb_lib.c: fix build without TLS

x1sc0:
I'm getting ma_pvio_tls_verify_server_cert and reset_tls_error being undeclared


9EOR9:
You can't build without TLS anymore. E.g. for various authentication plugins hashing is required which is provided by OpenSSL/GnuTLS.

PR #288, opened 2025-07-02
Fix miscellaneous typos

qobood:


qobood:
There is a function "test_cuted_rows", which I'd like to change to "test_cut_rows", although I don't know if it is used externally. Should it be just left as is to avoid any issues?

9EOR9:
A few comments:

* Could you please use the 3.3 branch (we will then merge into 3.4)?

* Please don’t fix external/zlib. If you want to correct typos in zlib, you should file a pull request directly with the zlib project.We update the zlib sources frequently, so corrections would get lost.

* The man pages are autogenerated; typos need to be fixed in the wiki. Since we are currently moving to GitHub Books, I will forward your man page corrections to the documentation team so they won’t be forgotten.

qobood:
Hi @9EOR9!
I've made new PR  for 3.3 branch, pulled out external/zlib/ and man/
https://github.com/mariadb-corporation/mariadb-connector-c/pull/289
Thank you!

PR #291, opened 2025-07-31
Fix some more typos

qobood:


qobood:
Hi @9E0R9!
I have two questions:
There is a typo in variable name (cssl) in libmariadb/mariadb_async.c below #ifdef HAVE_TLS_FIXME. This part hasn't been touched since 2016, should it still be changed?
And another question - while testing I've bumped into a client crash because of lack of checks for NULL in mariadb_lib.c. I've submitted the bugreport:
https://jira.mariadb.org/browse/MDEV-37034
and it got closed. I'd still like to add appropriate checks. What do you think?

grooverdan (approved):

PR #292, opened 2025-08-15
CONC-774 fixup: Always emit BTI J

dr-m:
In commit e2e5113a530897d8aedc4ad1d65c5a99a926b9cc (#276) an incorrect numeric encoding was added that is for the instruction `BTI`, not `BTI J`.

According to https://www.spinics.net/lists/arm-kernel/msg1041915.html the two most significant bits of the least significant octet of the instruction word for `BTI` are [`J`][`C`] flags.

Because a newer version of GCC could emit code for an older assembler, we had better always use the numeric encoding on GCC. In clang and derivatives, the assembler is built-in.

### DIFF
--- a/libmariadb/ma_context.c
+++ b/libmariadb/ma_context.c
@@ -638,11 +638,19 @@ my_context_destroy(struct my_context *c)
 
 #ifdef MY_CONTEXT_USE_AARCH64_GCC_ASM
 
-#if (defined(__clang__) && (__clang_major__ >= 9)) ||                   \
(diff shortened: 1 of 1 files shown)


knielsen (approved):
I'm not familiar with the instruction encodings here, I trust Marko's analysis here.

Generally I'm not in favour of these kinds of "security" bloats, and even if they are wanted sometimes ideally the extra instruction overhead should only be added if the code is compiled with this explicitly enabled (like the compiler does). But it's not important enough to spend any more effort on, this should be fine.

PR #294, opened 2025-09-18
mariadb_config: not for AIX

grooverdan:
AIX no longer provide the getoptlong compatible
library for mariadb_config. AIX applications
depending on this will need to use an alternate
method of determining headers and libraries of
C/C.

Tested on AIX-7.3 in MariaDB Server 10.11 submodule

### DIFF
--- a/mariadb_config/CMakeLists.txt
+++ b/mariadb_config/CMakeLists.txt
@@ -44,23 +44,22 @@ IF(UNIX AND NOT APPLE)
   ENDIF()
 ENDIF()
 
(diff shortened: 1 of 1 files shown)


vuvova:
I think if AIX doesn't provide a compatible getopt library, then whoever builds mariadb on AIX should, we can install it on the builder. We don't do rpm/deb AIX packages with proper dependency tracking, so we can require getopt.h to exist and provide the necessary functionality and how it'll be achieved is not the problem of our builds.

_I basically suggest to install this header-only getopt.h implementation on our AIX builder_

PR #295, opened 2025-12-01
mark RISC-V64 as 64-bit architecture

hrw:


cmake/install.cmake
```diff
@@ -76,7 +76,7 @@ SET(LIBMARIADB_STATIC_DEFAULT "mariadbclient")
 # RPM layout
 #
 SET(INSTALL_BINDIR_RPM "bin")
-IF((CMAKE_SYSTEM_PROCESSOR MATCHES "x86_64" OR CMAKE_SYSTEM_PROCESSOR MATCHES "ppc64" OR CMAKE_SYSTEM_PROCESSOR MATCHES "ppc64le" OR CMAKE_SYSTEM_PROCESSOR MATCHES "aarch64" OR CMAKE_SYSTEM_PROCESSOR MATCHES "s390x") AND CMAKE_SIZEOF_VOID_P EQUAL 8)
```
grooverdan, line 79: `if(CMAKE_SIZEOF_VOID_P EQUAL 8)` would involve less processor enumeration.

grooverdan (approved):
otherwise looks fine to me.

cmake/install.cmake
hrw: I wanted to not involve bigger changes so that's why I added one more architecture to the list.

This way it is up to project maintainers which way they want to go further.

ottok:
This has been in Fedora since Dec 1st: https://src.fedoraproject.org/rpms/mariadb11.8/blob/rawhide/f/mark-RISC-V64-as-64-bit-architecture.patch

PR #297, opened 2026-01-01
MDEV-35254 Make iterations count configurable in PARSEC plugin

deo002:
This patch changes the maximum permissible value of the iteration factor in the auth string.

### DIFF
--- a/plugins/auth/parsec.c
+++ b/plugins/auth/parsec.c
@@ -47,6 +47,7 @@
 #define ED25519_KEY_LENGTH        32
 #define PBKDF2_HASH_LENGTH        ED25519_KEY_LENGTH
 #define CLIENT_RESPONSE_LENGTH    (CHALLENGE_SCRAMBLE_LENGTH + ED25519_SIG_LENGTH)
+#define PARSEC_ITER_FACTOR_MAX    20
 
 struct Passwd_in_memory
 {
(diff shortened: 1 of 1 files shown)


plugins/auth/parsec.c
```diff
@@ -47,6 +47,7 @@
 #define ED25519_KEY_LENGTH        32
 #define PBKDF2_HASH_LENGTH        ED25519_KEY_LENGTH
 #define CLIENT_RESPONSE_LENGTH    (CHALLENGE_SCRAMBLE_LENGTH + ED25519_SIG_LENGTH)
+#define PARSEC_ITER_FACTOR_MAX    9
```
vuvova: this isn't in sync with the server. Why the discrepancy? Make it 9 in the server too. Or make it 62 here.

deo002: Yes, I had changed the max value in server, forgot about this one.

vuvova (approved):

plugins/auth/parsec.c
```diff
@@ -47,6 +47,7 @@
 #define ED25519_KEY_LENGTH        32
 #define PBKDF2_HASH_LENGTH        ED25519_KEY_LENGTH
 #define CLIENT_RESPONSE_LENGTH    (CHALLENGE_SCRAMBLE_LENGTH + ED25519_SIG_LENGTH)
+#define PARSEC_ITER_FACTOR_MAX    21
```
vuvova: make it 20, please (because `PKCS5_PBKDF2_HMAC()` takes a signed int, and `1 << 31` is too large).

then I'll merge.

vuvova (changes requested):

deo002:
Done!

vuvova (approved):

PR #298, opened 2026-01-06
MDEV-38550: add LENENC support for COM_CHANGE_USER

rophy:
See https://github.com/MariaDB/server/pull/4534

Add CLIENT_PLUGIN_AUTH_LENENC_CLIENT_DATA support to send_change_user_packet()
to allow passwords longer than 255 bytes in COM_CHANGE_USER packets.

Previously, send_change_user_packet() only supported single-byte length
encoding (max 255 bytes), while send_client_reply_packet() supported LENENC.
This caused mysql_change_user() to fail with long passwords when the
connection used cleartext auth directly (without auth plugin switching).

Note: This fix requires a corresponding server-side fix in
parse_com_change_user_packet() to parse LENENC-encoded passwords.

### DIFF
--- a/unittest/libmariadb/connection.c
+++ b/unittest/libmariadb/connection.c
@@ -368,92 +368,6 @@ static int test_change_user(MYSQL *mysql)
   return OK;
 }
 
-/*
-  MDEV-38550: Test mysql_change_user() with long password (>255 bytes)
-
-  Uses PAM auth with cleartext plugin so the raw password is sent as auth data
-  in the COM_CHANGE_USER packet. A 260-byte password produces 261 bytes of
-  auth data (including null terminator), which requires LENENC encoding
-  instead of single-byte length.
-
-  Requires:
-  - Server with MDEV-38550 fix (MariaDB/server#4534)
-  - auth_pam plugin loaded with pam-use-cleartext-plugin
-  - PAM service 'mariadb_mdbtest' configured with pam_permit
-*/
-static int test_change_user_long_pw(MYSQL *mysql)
-{
-  char long_pw[261];
-  int rc;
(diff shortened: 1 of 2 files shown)


unittest/libmariadb/connection.c
```diff
@@ -368,6 +368,92 @@ static int test_change_user(MYSQL *mysql)
...
+  char long_pw[261];
+  int rc;
+  MYSQL *conn;
+  const char *user= "mdev38550_pam_user";
+
+  SKIP_MAXSCALE;
+  SKIP_SKYSQL;
+
+  /* Check if PAM auth plugin is available */
+  rc = mysql_query(mysql, "INSTALL SONAME 'auth_pam'");
```
vuvova: why not the cleartext server plugin?

Practically, I suspect the test you've added will ever be run, so you can as well remove it and rely on server test.

vuvova (changes requested):
see comments inl

unittest/libmariadb/connection.c
rophy: Removed the test case. `cleartext_plugin_server` lives in `auth_test_plugin.so` (tagged `COMPONENT Test` in CMake) and isn't shipped in the `mariadb:10.11` image or `mariadb-test` apt package, so the test would be skipped in CI. Happy to restore it if there's a standard way to load `auth_test_plugin.so` in connector-c CI.

vuvova: I think server-only test is sufficient, if connector/c will break this again, the server test will fail

vuvova (approved):
Looks good. ~Please, squash~ Actually, nevermind

PR #299, opened 2026-01-09
MDEV-38499: cmake and compile errors on MacOSX when compiling mariadb

gkodinov:
Replace deprecated cmake function calls to avoid warnings

Replaced the deprecated exec_program calls with exeecute_process() This removed the deprecation cmake warnings.


vaintroub (commented):
Fix typo in commit comment "exeecute_process", correctly "execute_process".
Otherwise ok

vaintroub (approved):

PR #300, opened 2026-01-26
[CONC-806] SCHANNEL Async Support and SSL Buffer Management Optimization

rusher:
see https://jira.mariadb.org/browse/CONC-806 description


rusher:
not the best implementation. closing it

PR #302, opened 2026-03-12
Fix ninja build for DEB layout

FooBarrior:
In DEB layout, mariadbclient is named libmariadb.a

In install.cmake:
SET(LIBMARIADB_STATIC_DEB "mariadb")

This causes a duplicate target for libmariadb.a, and moreover, a symlink recipe with reference to already existing file.

Make builds don't report a dupliate problem, and besides, "file already exists" is not reported somehow.
Ninja build reports error.

The only check that has to be done is comparison
against INSTALL_LAYOUT, because LIBMARIADB_STATIC_NAME is unconditionally set to the layout's value:
SET(LIBMARIADB_STATIC_NAME ${LIBMARIADB_STATIC_${INSTALL_LAYOUT}})

### DIFF
--- a/libmariadb/CMakeLists.txt
+++ b/libmariadb/CMakeLists.txt
@@ -490,7 +490,10 @@ IF(WITH_MYSQLCOMPAT)
   ENDIF()
 ENDIF()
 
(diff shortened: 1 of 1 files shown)


FooBarrior:
The macos failure is 
```
14: # Error: TLS/SSL error: Operation timed out (60)
```
Seems unrelated to the patch

PR #305, opened 2026-03-29
CONC-813, CONC-814 Fix UBSan errors

FaramosCZ:
sint4korr (non-x86 path): the expression
`(int16)(A)[3] << 24` causes signed integer overflow when byte value >= 128, since 128<<24 exceeds INT32_MAX. Redefine as `(int32) uint4korr(A)` to compute in unsigned arithmetic, matching the server (my_byteorder.h:137-140). The next line sint8korr already uses this same pattern: `(longlong) uint8korr(A)`.

convert_from_long (MYSQL_TYPE_DOUBLE, MYSQL_TYPE_FLOAT): casting double/float back to ulonglong/longlong for truncation detection is undefined when the float value is outside the representable integer range. For example, `(double)ULONGLONG_MAX` rounds up to 2^64 in IEEE 754 (not exactly representable with 53-bit mantissa), so `(ulonglong)dbl` overflows. Similarly for LONGLONG_MAX. Add range checks before the cast using `>=` for upper bounds (since the double rounds up past the integer max) and `<` for LONGLONG_MIN (exactly representable as -2^63). Out-of-range values set the truncation error flag directly, short-circuiting the dangerous cast via `||` evaluation. This matches the server pattern in sql/field.cc and sql/sql_select.cc (double_to_ulonglong macro).

### DIFF
--- a/include/ma_global.h
+++ b/include/ma_global.h
@@ -832,10 +832,7 @@ do { doubleget_union _tmp; \
 				  (((uint32) (uchar) (A)[2]) << 16) |\
 				  (((uint32) (uchar) (A)[1]) << 8) | \
 				  ((uint32) (uchar) (A)[0])))
-#define sint4korr(A)	(int32) (((int32) ((uchar) (A)[0])) +\
-				(((int32) ((uchar) (A)[1]) << 8)) +\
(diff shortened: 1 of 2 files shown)


FaramosCZ:
Requesting review: @9EOR9

PR #307, opened 2026-04-07
CONC-817: CMake cleanup: fix variable quoting bugs, remove dead code

FaramosCZ:
Fix several small CMake issues: unquoted variable expansions that can mis-parse, a redundant conditional, incorrect variable names, and removal of dead/unused variables and code blocks.

*Bugs:*

- Remove redundant nested IF(MSVC) inside an outer IF(MSVC) block (CMakeLists.txt) — inner condition is always true

- Quote unquoted variable expansions in IF() conditions that can
  break when the value is empty or multi-word:
  ${IS_SO} and ${LIB_PREFIX} in mariadb_config/CMakeLists.txt,
  ${v}_FILE_VERSION in CMakeLists.txt,
  ${CC_TARGET_COMPILE_OPTIONS} in cmake/plugins.cmake

- Fix ${ZLIB_LIBRARY} (singular, never set by CMake's FindZLIB)
  to ${ZLIB_LIBRARIES} (the standard variable) in CMakeLists.txt
  and libmariadb/CMakeLists.txt — standalone builds silently
  dropped zlib from the link line; the server masked this with
  a compatibility shim in cmake/zlib.cmake

*Dead code removal:*

- ADD_DEFINITIONS(${LIBMARIADB_PLUGIN_DEFS}) — variable never
  populated by the plugin registration system (since 50d48e91)

- ${LIBBIND} — SEARCH_LIBRARY call removed in 97a6aeaf (2016)
  but references left behind

- PLUGINS_OFF — accumulated but never read, also lacks
  PARENT_SCOPE so is function-local

- PLUGIN_EXTRA_FILES — set but never referenced; the file it
  names is already compiled into libmariadb directly

- GSSAPI_FLAVOR_MIT — set to empty, never used; only
  GSSAPI_FLAVOR (without _MIT) is consumed

- Commented-out CLIENT_DOCS install block — CLIENT_DOCS no
  longer set since the manpage revert in 17f2d1fe (2024);
  documentation now handled via ADD_SUBDIRECTORY(man)

- Duplicate bundled-zlib IF/ELSE block in
  libmariadb/CMakeLists.txt — first block defines ZLIB_SOURCES
  but never adds them to LIBMARIADB_SOURCES; only the second
  block is functional


FaramosCZ:
rebased

FaramosCZ:
Ready for review @9EOR9

PR #310, opened 2026-04-24
Windows: make Schannel async read/write cooperate with non-blocking client API

Alex-Tsvetanov:
## Summary

This PR carries two commits that together close the last visible gap in the Windows Connector/C async API — the one commit `9644f527` ("Skip async test on Windows") documented as "Schannel implementation doesn't support async mode yet":

- **`Windows: make Schannel async read/write cooperate with non-blocking client API`** (`7cab1160`) — the functional fix.
- **`test_conc760: skip (not fail) when server doesn't expose @@named_pipe`** (`7e7e3595`) — a pre-existing test bug surfaced while getting `ctest` fully green; harmless on its own but blocks a clean 21/21 run against a non-Windows server.

With both commits, `unittest/libmariadb/async.c` passes 5/5 on Windows over Schannel TLS, and the file is unskipped in `unittest/libmariadb/CMakeLists.txt` as the regression guard.

## Root cause

Two compounding holes in libmariadb 3.4 on Windows + Schannel:

1. **`libmariadb/ma_pvio.c`** compile-time excluded the TLS async dispatcher for Schannel builds: `#if defined(HAVE_TLS) && !defined(HAVE_SCHANNEL)`. With `HAVE_SCHANNEL` defined, the async path fell through to `ma_pvio_read_async` / `_write_async` → raw socket I/O while the wire carried TLS ciphertext. The MySQL packet parser interpreted the ciphertext as a malformed header, and either bailed or caused the server to close the connection as a protocol violation. That is why reproducers saw `Lost connection to server during query` even though the server was healthy.

2. Even if the dispatcher reached Schannel, **`ma_schannel_read_decrypt` / `_write_encrypt` reach the socket through `pvio->methods->read/write`**, which on Windows handle `WSAEWOULDBLOCK` by calling `pvio_socket_wait_io_or_timeout` → `select()`. If that happens inside the async fiber, the fiber's OS thread blocks — and because Windows fibers cooperatively multiplex on a single OS thread, the application fiber never runs, so the app can never `select()` its own fd set, so we deadlock until the server-side timeout. This is the symptom documented by commit `9644f527` ("Schannel implementation doesn't support async mode yet").

Symmetric evidence: with TLS disabled via `MARIADB_TLS_DISABLE_PEER_VERIFICATION=1` (which flips the implicit `use_ssl=1` off in modern 3.4), both zero-coroute reproducers succeed on Windows immediately. With TLS on they fail. That scopes the defect precisely to the Schannel async path.

## Fix (4 files, +83 / −11)

| File | Change |
|---|---|
| `plugins/pvio/pvio_socket.c` | New `pvio_socket_wait_or_yield` — when `IS_PVIO_ASYNC_ACTIVE(pvio)` is true, set `events_to_wait_for` on the async context and call `my_context_yield` (cooperative yield to the app fiber) instead of blocking in `select()`. The two `pvio_socket_read` / `_write` call sites use the new helper; the unmodified `pvio_socket_wait_io_or_timeout` stays available for internal connect / keepalive. |
| `libmariadb/ma_pvio.c` | Drop `&& !defined(HAVE_SCHANNEL)` from the two TLS-async dispatch guards so `ma_tls_read_async` / `ma_tls_write_async` are called for Schannel builds. |
| `libmariadb/secure/schannel.c` | Provide `ma_tls_read_async` / `ma_tls_write_async` as trivial delegators to the existing sync `ma_tls_read` / `ma_tls_write`. The Schannel decrypt/encrypt helpers already reach the socket via `pvio->methods->read/write`, and those now yield cooperatively (step 1), so no per-layer async state machine is needed. |
| `unittest/libmariadb/CMakeLists.txt` | Drop `IF(NOT WIN32)` around `SET(API_TESTS ${API_TESTS} "async")`. `async.c` is now the regression guard for this fix on Windows. |

Bonus `test_conc760` fix (`unittest/libmariadb/connection.c`): the pre-existing test passed the wrong `MYSQL*` to `check_mysql_rc` and was unconditionally `FAIL`ing on any server that doesn't expose `@@named_pipe` (i.e. any non-Windows server — same problem in Connector/C CI against a Linux server). Now skips gracefully with a diagnostic; test-typo is fixed.

## Verification

Built this branch from source (no vcpkg, no binary package). Target server: stock `mariadb:latest` in docker on `localhost:3306`. Windows 11 + MSVC.

### Zero-coroute reproducers (included in `.claude-investigation/` on the fork branch but excluded from the PR diff)

| Reproducer | Before | After |
|---|---|---|
| `reproducer_A_async_connect.cpp` — canonical `mysql_real_connect_start` + plain `select()` over default Schannel TLS | `async connect FAILED: Lost connection to server at 'reading authorization packet'` | `async connect OK` → `GOT: 1` → `DONE OK` |
| `reproducer_B_async_query.cpp` — sync `mysql_real_connect` + manual `ioctlsocket(FIONBIO)` + `mysql_real_query_start/cont` | `query_cont status=0 err=1 msg=Lost connection to server during query` | `GOT: 1` → `DONE OK` |

### Upstream unit suite (`unittest/libmariadb`)

```
21/21 Test #21: async ............................   Passed   17.84 sec

100% tests passed, 0 tests failed out of 21
Total Test time (real) = 145.69 sec
```

Including the newly-unskipped `async.c`:

```
1..5
ok 1 - test_async
ok 2 - async1          # 100 iterations of connect_start/cont + query_start/cont + store_result_start/cont + close_start/cont
ok 3 - test_conc131
ok 4 - test_conc129
ok 5 - test_conc622
# Cipher in use: TLS_AES_256_GCM_SHA384
# Subject: CN=MariaDB Server
```

Cipher line confirms Schannel TLS is active end-to-end (client = Schannel; server = OpenSSL 3.0.13 in the docker image).

### Downstream verification

This branch was built and linked into the coroute ORM integration at `cpp-for-everything/WebFrame`'s `feature/orm-integration`. The coroute suite has its own Windows gate on the `AsyncMariaDB` reactor tests (previously `AsyncMySQLDB`, renamed along the way) that was flipped off for this experiment. Full DB suite against all four containers (postgres:16, mariadb:latest, redis:latest, mongo:latest): **860 assertions / 162 test cases, all green.** Delta vs. the prior Windows baseline of 837/159: **+3 test cases** — the three reactor CRUD tests that were `#if !defined(_WIN32)` gated when the Windows async path was broken.

## Platform testing scope

Tested on Windows 11 + MSVC 18 Insiders against docker `mariadb:latest`. Not tested on Linux/macOS here, but the `pvio_socket_wait_or_yield` change is effectively inert on those platforms:

- Non-Windows + no TLS async goes through `pvio_socket_async_read` (with `MSG_DONTWAIT`), not `pvio_socket_read`'s EWOULDBLOCK path that the helper gates.
- Non-Windows + OpenSSL/GnuTLS async yields inside the existing `ma_tls_*_async` helpers via `SSL_ERROR_WANT_READ`/`_WANT_WRITE` — again never reaching `pvio_socket_read`'s loop.

So the helper only changes behavior on Windows + Schannel + async-active, which is exactly the code path being fixed. Happy to re-verify against your Linux CI once this PR is open.

## Test plan

- [x] `unittest/libmariadb` full suite (`ctest -C Release`) 21/21 green on Windows vs docker `mariadb:latest`
- [x] `unittest/libmariadb/async.c` unskipped and green on Windows (5/5)
- [x] Downstream linked build + exercise of the async API from a real application (coroute ORM) green on Windows — 9 test cases / 49 assertions specifically for the MariaDB reactor path
- [ ] Your Linux CI — passes unchanged (change is inert on non-Windows paths; happy to debug if not)

### DIFF
--- a/libmariadb/ma_pvio.c
+++ b/libmariadb/ma_pvio.c
@@ -224,7 +224,7 @@ ssize_t ma_pvio_read(MARIADB_PVIO *pvio, uchar *buffer, size_t length)
   if (IS_PVIO_ASYNC_ACTIVE(pvio))
   {
     r=
-#if defined(HAVE_TLS) && !defined(HAVE_SCHANNEL)
+#if defined(HAVE_TLS)
       (pvio->ctls) ? ma_tls_read_async(pvio, buffer, length) :
 #endif
                     (ssize_t)ma_pvio_read_async(pvio, buffer, length);
@@ -352,7 +352,7 @@ ssize_t ma_pvio_write(MARIADB_PVIO *pvio, const uchar *buffer, size_t length)
   if (IS_PVIO_ASYNC_ACTIVE(pvio))
   {
     r=
-#if defined(HAVE_TLS) && !defined(HAVE_SCHANNEL)
+#if defined(HAVE_TLS)
        (pvio->ctls) ? ma_tls_write_async(pvio, buffer, length) :
 #endif
                       ma_pvio_write_async(pvio, buffer, length);
--- a/libmariadb/secure/schannel.c
+++ b/libmariadb/secure/schannel.c
@@ -666,7 +666,7 @@ retry:
 }
 
 ssize_t ma_tls_write(MARIADB_TLS *ctls, const uchar* buffer, size_t length)
-{ 
+{
   MARIADB_PVIO *pvio= ctls->pvio;
   ssize_t rc, wlength= 0;
   ssize_t remain= length;
@@ -679,6 +679,32 @@ ssize_t ma_tls_write(MARIADB_TLS *ctls, const uchar* buffer, size_t length)
     remain-= rc;
   }
   return length;
+}
+
+/*
+  Schannel async read/write.
+
+  ma_schannel_read_decrypt and ma_schannel_write_encrypt reach the socket via
+  pvio->methods->read / ->write. When the connection is executing inside an
+  async fiber, the socket plugin (pvio_socket_wait_or_yield) cooperatively
+  yields back to the app fiber on EWOULDBLOCK instead of blocking the OS
+  thread in select(). That makes the ordinary sync TLS read/write paths
+  async-safe with no per-record state machine here: we simply delegate to the
+  existing ma_tls_read / ma_tls_write, and any would-block during a handshake
+  record, decrypt, or encrypt round-trip is handled one layer deeper.
+*/
+ssize_t ma_tls_read_async(MARIADB_PVIO *pvio, const uchar *buffer, size_t length)
+{
+  if (!pvio || !pvio->ctls)
+    return -1;
+  return ma_tls_read(pvio->ctls, buffer, length);
+}
+
+ssize_t ma_tls_write_async(MARIADB_PVIO *pvio, const uchar *buffer, size_t length)
+{
+  if (!pvio || !pvio->ctls)
+    return -1;
+  return ma_tls_write(pvio->ctls, buffer, length);
 }
 
 /* {{{ my_bool ma_tls_close(MARIADB_PVIO *pvio) */
--- a/plugins/pvio/pvio_socket.c
+++ b/plugins/pvio/pvio_socket.c
@@ -109,13 +109,61 @@ my_bool pvio_socket_is_alive(MARIADB_PVIO *pvio);
 my_bool pvio_socket_has_data(MARIADB_PVIO *pvio, ssize_t *data_len);
 int pvio_socket_shutdown(MARIADB_PVIO *pvio);
 
-static int pvio_socket_init(char *unused1, 
-                           size_t unused2, 
-                           int unused3, 
+static int pvio_socket_init(char *unused1,
+                           size_t unused2,
+                           int unused3,
                            va_list);
 static int pvio_socket_end(void);
 static ssize_t ma_send(my_socket socket, const uchar *buffer, size_t length, int flags);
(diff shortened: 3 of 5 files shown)


Alex-Tsvetanov:
The previous CI run failed only in the `server 11.4 test suite` job, on `main.group_by` — a non-deterministic UNION ordering unrelated to this PR (which only touches Windows Schannel async I/O):

```
CURRENT_TEST: main.group_by
--- mysql-test/main/group_by.result
+++ mysql-test/main/group_by.reject
@@ -2636,8 +2636,8 @@
 SET sql_mode='ONLY_FULL_GROUP_BY';
 SELECT 1 AS test UNION SELECT 2 AS test ORDER BY test IS NULL ASC;
 test
-1
 2
+1
 SET sql_mode='';
```

The ORDER BY key (`test IS NULL`) evaluates to `0` for every row, so the row order is unspecified. The same job failed on the scheduled run over untouched 3.4 on 2026-04-19 ([run 24619796993](https://github.com/mariadb-corporation/mariadb-connector-c/actions/runs/24619796993/job/71988293365)) — so it's a pre-existing flake in the upstream server suite, not a regression here.

Upstream fix submitted at [MariaDB/server#4991](https://github.com/MariaDB/server/pull/4991) (adds `--sorted_result` so mtr sorts the client output before diffing; no `.result` change).

Pushing an empty commit to retrigger CI.

PR #311, opened 2026-05-14
Fix redefining `bool` when including `ma_global.h` in C++

ParadoxV5:
If `HAVE_BOOL` is not defined (by default), the C++ check didn’t matter because of the `||` operator.


ParadoxV5:
3.1 EOL? Not said anywhere

9EOR9:
3.1 is EOL. Please use 3.3 - it will be merged afterwards into 3.4

ParadoxV5:
Done, @9EOR9, just needs a rebase with squash.

By the way, does Connector/C document EOLs / support terms somewhere?

PR #312, opened 2026-05-19
MDBF-1214 Use HEAD for Git source package archives

RazvanLiviuVarzaru:
Buildbot source package builds can run from a shallow detached checkout created by fetching a specific commit and checking out FETCH_HEAD. In that state, git show-branch may not report a branch name, leaving GIT_BRANCH empty and causing git archive to fail.

Archive HEAD directly instead. This packages the commit that CMake is building, regardless of whether it was checked out from a branch, tag, or detached CI commit.


grooverdan (approved):
good durable simplification 🍰

PR #315, opened 2026-06-30
CONC-830: add missing MARIADB_CLIENT_PLUGIN_EXPORT to zstd plugin

FaramosCZ:
MDEV-37527 added '-fvisibility=hidden' for plugin builds and the 'MARIADB_CLIENT_PLUGIN_EXPORT' macro to explicitly export '_mysql_client_plugin_declaration_' from each plugin. The zstd compression plugin was missed, leaving it with zero exported symbols and making it impossible to load dynamically via 'dlsym()'.


FaramosCZ:
fixed as https://github.com/mariadb-corporation/mariadb-connector-c/commit/9f33db7aff00c481e27379c3e347cedcbd6df3cb

PR #317, opened 2026-07-01
CONC-821 Regression: mysql_stmt_bind_result zeroes *length

grooverdan:
CONC-812 needed a length after a NULL was returned.

What is required for compatibiliy is for the length to always be set.

Adding 5fe4d03f26045d1215e894d16054935ddd65ecc0 in CONC-812 this was the correct fix. For readability this is now in the default statement after all the fixed type lengths have been populed.

   if (mysql_ps_fetch_functions[stmt->fields[i].type].pack_len < 0)
      *stmt->bind[i].length= stmt->bind[i].length_value= 0;

Evidently MSAN noticed that in some cases there wasn't a value set. This resulted in the commit. 40806da977ae073e13e64e3b68dc42e3f63020b0

This left variable length fields reset to 0.

To ensure that MSAN remains happy with a SELECT NULL result, in mthd_stmt_fetch_to_bind, when the pack_len is 0, only the case for MYSQL_TYPE_NULL, set the length to 0.

In mysql_stmt_bind_result, as pack_length is set for fixed type we only need to set the length for the temporal types in the switch statement.

### DIFF
--- a/libmariadb/ma_stmt_codec.c
+++ b/libmariadb/ma_stmt_codec.c
@@ -1131,6 +1131,7 @@ void ps_fetch_datetime(MYSQL_BIND *r_param, const MYSQL_FIELD * field,
   MYSQL_TIME *t= (MYSQL_TIME *)r_param->buffer;
   unsigned int len= net_field_length(row);
 
+  *r_param->length= sizeof(MYSQL_TIME);
   switch (r_param->buffer_type) {
     case MYSQL_TYPE_DATETIME:
     case MYSQL_TYPE_TIMESTAMP:
--- a/libmariadb/mariadb_stmt.c
+++ b/libmariadb/mariadb_stmt.c
@@ -445,7 +445,11 @@ int mthd_stmt_fetch_to_bind(MYSQL_STMT *stmt, unsigned char *row, ulong length)
         stmt->bind[i].u.row_ptr= NULL;
         if (!stmt->bind[i].length)
           stmt->bind[i].length= &stmt->bind[i].length_value;
-        if (mysql_ps_fetch_functions[stmt->fields[i].type].pack_len < 0)
+	/*
+	  include pack_len 0 (MYSQL_TYPE_NULL as this may have
+	  previously been a MYSQL_TYPE_STRING at the bind phase.
+	*/
+        if (mysql_ps_fetch_functions[stmt->fields[i].type].pack_len <= 0)
           *stmt->bind[i].length= stmt->bind[i].length_value= 0;
       }
     } else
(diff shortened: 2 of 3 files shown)


FaramosCZ:
Thanks for working on this, @grooverdan.

**Overall approach looks correct** — the original `else { *length = 0; }` branch in `mysql_stmt_bind_result()` was wrong because it zeroed `*length` for both string/blob types (fine) and temporal types (breaks the contract: temporal columns have a known fixed length of `sizeof(MYSQL_TIME)`).

One issue I'd like to flag:

### Temporal `*length` left indeterminate between bind and first fetch

After this PR, `mysql_stmt_bind_result()` sets `*length` only when `pack_len >= 0` (fixed-size numeric types). Temporal types (`TIME`, `DATE`, `DATETIME`, `TIMESTAMP`) have `pack_len = MYSQL_PS_SKIP_RESULT_W_LEN` (= −1), so they fall through with **no assignment**. The `sizeof(MYSQL_TIME)` value is written later in `ps_fetch_datetime()` — but only when the first row is actually fetched.

This means between `mysql_stmt_bind_result()` and `mysql_stmt_fetch()`, `*bind.length` for temporal columns is **indeterminate**. If the application inspects `*length` between these two calls (e.g. to pre-allocate buffers, or in a loop that processes bind metadata before the first fetch), it will read uninitialized or stale data.

The [MySQL 8.4 docs](https://dev.mysql.com/doc/c-api/8.4/en/c-api-prepared-statement-data-structures.html) describe `*length` as:

> *"For fixed-size types [the actual length] is the size of the data type."*

Temporal types fall in this category — they always occupy `sizeof(MYSQL_TIME)` regardless of the wire format. The pre-regression code (the commented-out switch block that this PR removes) set `sizeof(MYSQL_TIME)` at bind time for exactly this reason.

MariaDB Docs are not specific enough about temporal types, and  when `*length` gets populated (bind-time vs fetch-time)

I would suggest adding a `sizeof(MYSQL_TIME)` assignment for temporal types in `mysql_stmt_bind_result()` alongside the fixed numerics — not just in the fetch path. Something like:

```c
if (mysql_ps_fetch_functions[stmt->bind[i].buffer_type].pack_len >= 0)
    *stmt->bind[i].length = stmt->bind[i].length_value =
        mysql_ps_fetch_functions[stmt->bind[i].buffer_type].pack_len;
else switch(stmt->bind[i].buffer_type) {
    case MYSQL_TYPE_TIME:
    case MYSQL_TYPE_DATE:
    case MYSQL_TYPE_DATETIME:
    case MYSQL_TYPE_TIMESTAMP:
        *stmt->bind[i].length = stmt->bind[i].length_value = sizeof(MYSQL_TIME);
        break;
    default:
        break;
}
```

This matches the old switch-based approach, keeps MSAN happy (all `*length` slots are written or intentionally skipped), and ensures temporal columns report `sizeof(MYSQL_TIME)` immediately after bind — not just after the first fetch.

**Additional context**: The regression from `40806da9` was already hit by downstream consumers:
- [perl-DBD-MariaDB](https://metacpan.org/pod/DBD::MariaDB) broke in Fedora 45 with mariadb-connector-c 3.4.9 (`mariadb_st_internal_execute()` [reads `*length` right after bind](https://github.com/gooddata/DBD-MariaDB/blob/v1.24/dbdimp.c#L2541) to compute fetch offsets)
- [CONCPP-152](https://jira.mariadb.org/browse/CONCPP-152), [diesel#5091](https://github.com/diesel-rs/diesel/discussions/5091)

The attempt for a [downstream patch in Fedora](https://src.fedoraproject.org/fork/mschorm/rpms/mariadb-connector-c/c/03b2e1d2171f33529383f462e64fb0c8fe0812db) that restores the switch-based assignment (equivalent to what I'm suggesting above), resolves the perl-DBD-MariaDB FTBFS.

libmariadb/ma_stmt_codec.c
```diff
@@ -1131,6 +1131,7 @@ void ps_fetch_datetime(MYSQL_BIND *r_param, const MYSQL_FIELD * field,
   MYSQL_TIME *t= (MYSQL_TIME *)r_param->buffer;
   unsigned int len= net_field_length(row);
 
+  *r_param->length= sizeof(MYSQL_TIME);
```
9EOR9, line 1134: No need to set *r_param->length  here - we shouldn't set length for any fixed sized types (numerical and temporal types) in fetch functions at all. 

Please also note that MYSQL_TYPE_YEAR is a temporal type which length is 2 and not sizeof(MYSQL_TIME).

libmariadb/mariadb_stmt.c
```diff
@@ -445,7 +445,11 @@ int mthd_stmt_fetch_to_bind(MYSQL_STMT *stmt, unsigned char *row, ulong length)
         stmt->bind[i].u.row_ptr= NULL;
         if (!stmt->bind[i].length)
           stmt->bind[i].length= &stmt->bind[i].length_value;
-        if (mysql_ps_fetch_functions[stmt->fields[i].type].pack_len < 0)
+	/*
+	  include pack_len 0 (MYSQL_TYPE_NULL as this may have
+	  previously been a MYSQL_TYPE_STRING at the bind phase.
+	*/
+        if (mysql_ps_fetch_functions[stmt->fields[i].type].pack_len <= 0)
```
9EOR9, line 452: Why <= 0 instead of < 0 ? MYSQL_TYPE_NULL has a pack_len of 0

We also need to exclude temporal types, e.g.

```
        if (mysql_ps_fetch_functions[stmt->fields[i].type].pack_len < 0 &&
           is_temporal_type(stmt->fields[i].type) == 0)
          *stmt->bind[i].length= stmt->bind[i].length_value= 0;

```
helper function:
```
static inline my_bool is_temporal_type(enum enum_field_types type)
{
  if (type == MYSQL_TYPE_TIME ||
      type == MYSQL_TYPE_TIMESTAMP ||
      type == MYSQL_TYPE_DATE ||
      type == MYSQL_TYPE_DATETIME ||
      type == MYSQL_TYPE_YEAR)
    return 1;
  return 0;
}
```

9EOR9 (commented):
The initial implementation of prepared statements assumed only numeric types are considered fixed-size types. This assumption is incorrect; according to the MySQL documentation, temporal types are also classified as fixed-size (though the documentation lacks an explicit mention of MYSQL_TYPE_TIMESTAMP). Consequently, for all fixed-size types, we should initialize and set the length once inside mysql_stmt_bind_result(), and these length values should remain unchanged even when a NULL value is received during a fetch.

libmariadb/mariadb_stmt.c
grooverdan: Yes, null was intended - showed in `test_conc762` test case having 3 there after fetch.

9EOR9: As discussed on Slack, we don’t need to unset length when a NULL value is returned. An application must always check is_null to distinguish between NULL and an empty string. Also, libmysql does not change the length value when NULL is returned

9EOR9: This would require changing test_conc762 in ps_bugs, which explicitly checks for length=0

9EOR9:
As discussed with Daniel, fixed by commit 0fb69426d13c14cb6af5ceb0dd9ae9d2777abea8.

Narrat:
Is this now different handling for variable sized types between `mariadb` and `mysql`?
I tried this fix with vaultwarden, which uses internally diesel-rs and was also affected by the same commit reported in CONC-821. But sadly this fix doesn't seem to help in this case.

No one specifically asked: What about the other reported programs? Are there testbuilds or tests in general happening?

PR #318, opened 2026-07-03
Use CMAKE_SIZEOF_VOID_P EQUAL 8 for 64-bit detection

sunhaiyong1978:
Use CMAKE_SIZEOF_VOID_P EQUAL 8 for 64‑bit detection instead of
    maintaining an architecture list. This is cleaner and more maintainable.


cmake/install.cmake
```diff
@@ -76,7 +76,7 @@ SET(LIBMARIADB_STATIC_DEFAULT "mariadbclient")
 # RPM layout
 #
 SET(INSTALL_BINDIR_RPM "bin")
-IF((CMAKE_SYSTEM_PROCESSOR MATCHES "x86_64" OR CMAKE_SYSTEM_PROCESSOR MATCHES "ppc64" OR CMAKE_SYSTEM_PROCESSOR MATCHES "ppc64le" OR CMAKE_SYSTEM_PROCESSOR MATCHES "aarch64" OR CMAKE_SYSTEM_PROCESSOR MATCHES "s390x" OR CMAKE_SYSTEM_PROCESSOR MATCHES "riscv64") AND CMAKE_SIZEOF_VOID_P EQUAL 8)
+IF((CMAKE_SYSTEM_PROCESSOR MATCHES "x86_64" OR CMAKE_SYSTEM_PROCESSOR MATCHES "ppc64" OR CMAKE_SYSTEM_PROCESSOR MATCHES "ppc64le" OR CMAKE_SYSTEM_PROCESSOR MATCHES "aarch64" OR CMAKE_SYSTEM_PROCESSOR MATCHES "s390x" OR CMAKE_SYSTEM_PROCESSOR MATCHES "riscv64" OR CMAKE_SYSTEM_PROCESSOR MATCHES "loongarch64") AND CMAKE_SIZEOF_VOID_P EQUAL 8)
```
grooverdan: I think rather than support an ever growing list of 64bit hardware, just keep the last criteria of `CMAKE_SIZEOF_VOID_P EQUAL 8`

sunhaiyong1978: Good point, I'll simplify it to just use CMAKE_SIZEOF_VOID_P EQUAL 8.

PR #321, opened 2026-07-15
CONC-837: cmake 4.4 deprecation warning for the CMAKE_MINIMUM_REQUIRED() version

gkodinov:
Upped the minimum required version to 3.20 to avoid a deprecation warning with cmake 4.4

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,7 +2,7 @@
 
 # This is the LGPL libmariadb project.
 
-CMAKE_MINIMUM_REQUIRED(VERSION 3.12.0 FATAL_ERROR)
+CMAKE_MINIMUM_REQUIRED(VERSION 3.11 FATAL_ERROR)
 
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -2,7 +2,7 @@
 
 # This is the LGPL libmariadb project.
 
-CMAKE_MINIMUM_REQUIRED(VERSION 3.12.0 FATAL_ERROR)
+CMAKE_MINIMUM_REQUIRED(VERSION 3.20 FATAL_ERROR)
```
9EOR9: Hi Georgi,

we need to support also older non eoled platforms, which don't provide newer cmake releases:
Ubuntu 20.04 LTS ships with cmake 3.16, RHEL8 has cmake 3.11 or upstream 3.18, Debian bullseye cmake 3.18

gkodinov: 3.11 works.

